PR #12, opened 2023-12-07
Allow provisioning

richfitz:

PR #13, opened 2023-12-07
Allow fetching task results via the driver

richfitz:
~Merge after #12, as it contains those commits~

Fairly small changes, allows handling of task results, which is trivial for windows of course


weshinsley (approved):
Great, all makes sense

PR #14, opened 2023-12-08
Add support for environments

richfitz:
This PR implements the "environments" support we talked about this morning. I don't love the name, but we can change that easily enough. Basic CRUD methods implemented (create, read, update, delete along with list). Special treatment for "default" which always exists, and which can be overwritten when uninitialised - I think this is the most pain-free general use.


weshinsley (approved):
Great - we discussed together

PR #15, opened 2023-12-08
Allow fancy expression capture

richfitz:
A new task creation approach, similar to explicit but more magic.

~Merge after #14, contains those commits~


weshinsley (approved):
All good - discussed this morning

PR #16, opened 2023-12-08
Provisioning, library and bootstrapping improvements

richfitz:
* Use the library that we install into for jobs
* Use the general boostrap library on I:
* Add an approach for creating the bootstrap library


weshinsley (approved):
Written together and discussed

PR #20, opened 2023-12-13
Support for cancelling tasks

richfitz:
Allow task cancellation. We end up bulk-cancelling tasks, drivers support doing this and they'll do so in reverse order on the windows driver.

Also renamed "started" to "running" everywhere


tests/testthat/helper-hermod.R
```diff
@@ -92,3 +106,21 @@ clear_drivers <- function() {
     rm(list = "drivers", envir = cache)
   }
 }
+
+
+wait_until_status <- function(id, wanted, root = NULL, time = 5, poll = 0.05) {
```
weshinsley, line 111: Not sure this gets called?

richfitz: yeah, that was leftover from an alternative bit of testing that I removed - dropped now

PR #22, opened 2023-12-13
Drop hermod prefix from tasks

richfitz:
No features, just change in function prefixes - check the NAMESPACE really, then the rest follows from this.


weshinsley (approved):
Looks good

PR #23, opened 2023-12-13
Allow configuration at initialisation

richfitz:
Allows passing `driver = "windows"` to `hermod_init()` which reduces boilerplate on initialisation. Also made `hermod_configure` more chatty


weshinsley (approved):
Great - all looks straightforward!

PR #28, opened 2023-12-15
Forbid mapping the I drive

weshinsley:


drivers/windows/R/paths.R
```diff
@@ -5,8 +5,7 @@ windows_path <- function(name, path_local, path_remote, drive_remote) {
   assert_scalar_character(drive_remote)
 
   if (!grepl("^[A-Za-z]:$", drive_remote)) {
-    stop(sprintf("drive_remote must be of the form 'X:' (but was '%s')",
-                 drive_remote))
+    cli::cli_abort("drive_remote must be of the form 'X:' (but was '%s')")
```
richfitz, line 8:
```suggestion
    cli::cli_abort("drive_remote must be of the form 'X:' (but was '{drive_remote}')")
```

drivers/windows/R/paths.R
```diff
@@ -15,11 +14,20 @@ windows_path <- function(name, path_local, path_remote, drive_remote) {
...
 
   if (!file.exists(path_local)) {
-    stop("Local mount point does not exist: ", path_local)
+    cli::cli_abort(c("Local mount point does not exist.",
+                 i = "Please check the mount point: {path_local}."))
+  }
+
+  if (drive_remote == "I:") {
+    cli::cli_abort(c("You cannot use I: on a cluster job.",
+                 i = "I: is reserved for package provisioning.",
```
richfitz, line 29:
```suggestion
                 i = "I: is reserved for internal use.",
```

More mysterious :shipit:

richfitz (changes requested):

drivers/windows/R/paths.R
weshinsley: Thanks - how did I miss that...!

drivers/windows/R/paths.R
weshinsley: hipercow moves in mysterious ways, its wonders to perform.

richfitz (approved):

PR #29, opened 2023-12-18
Allow advertising of globals

richfitz:
This PR lets the user declare what variables we can rely on to exist within their created environment. There's a special value `TRUE` here which just builds the env and works it out

### DIFF
--- a/R/environment.R
+++ b/R/environment.R
@@ -55,7 +55,6 @@ hipercow_environment_create <- function(name = "default", packages = NULL,
   ## look at this later, but literally noone wants it!
   path <- file.path(root$path$environments, name)
   exists <- file.exists(path)
-  prev <- if (exists) readRDS(path) else NULL
 
   if (exists && identical(readRDS(path), ret)) {
     cli::cli_alert_info("Environment '{name}' is unchanged")
--- a/R/environment.R
+++ b/R/environment.R
@@ -17,6 +17,10 @@
 ##'   for some reason, just call `library` yourself within one of your
 ##'   `source` files.
 ##'
+##' @param globals Names of global objects that we can assume exist
+##'   within this environment.  This might include function
+##'   definitions or large data objects.
+##'
 ##' @param overwrite On environment creation, replace an environment
 ##'   with the same name.
 ##'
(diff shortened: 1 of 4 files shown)


R/environment.R
```diff
@@ -46,6 +55,8 @@ hipercow_environment_create <- function(name = "default", sources = NULL,
   ## look at this later, but literally noone wants it!
   path <- file.path(root$path$environments, name)
   exists <- file.exists(path)
+  prev <- if (exists) readRDS(path) else NULL
```
weshinsley, line 58: prev doesn't get used

tests/testthat/test-environment.R
```diff
@@ -1,21 +1,24 @@
 test_that("can construct an environment", {
-  env <- new_environment("foo", NULL, NULL)
+  env <- new_environment("foo", NULL, NULL, NULL)
```
weshinsley, line 2: (note to self) -  the root arg is left unset here, but never gets reffed because sources and globals are both NULL

tests/testthat/test-interface.R
```diff
@@ -144,7 +144,7 @@ test_that("can call provision", {
 
   hipercow_provision(root = path_here, show_log = FALSE)
   mockery::expect_called(mock_provision, 1)
-  environment <- new_environment("default", NULL, NULL)
+  environment <- new_environment("default", NULL, NULL, NULL)
```
weshinsley, line 147: Ditto - root unset, but safe

R/environment.R
weshinsley, line 58:
```suggestion
```

weshinsley (approved):

tests/testthat/test-environment.R
weshinsley, line 2:
```suggestion
  env <- new_environment("foo", NULL, NULL, NULL, NULL)
```

tests/testthat/test-interface.R
weshinsley, line 147:
```suggestion
  environment <- new_environment("default", NULL, NULL, NULL, NULL)
```

weshinsley (approved):

PR #32, opened 2023-12-19
Drop 'name' argument from windows_path

richfitz:
Fairly mechanical - this is now unused everywhere so just drop it.


weshinsley (approved):
Great - all straightforward

PR #33, opened 2023-12-20
Allow preventing driver load

richfitz:
This is fairly nasty set of changes, despite their modest size:

1. allow preventing a driver from loading while running on the cluster (be7a136)
2. fix some bugs that made me realise that the above bug was happening (6ae953c)
3. add a workflow to allow us to install the development sources of hipercow for use on the cluster easily (035bee1, 7c7ae34)

The first one was caused by `task_status()` trying to load the windows driver on the task as it was running on windows! I've put logic in to prevent this and an assertion that will fail nice and early should it happen again. We set an environment variable in the bootstrap that will prevent driver load.

The second one was a few different small bugs:

* we don't create the job library before setting `.libPaths()` to use it (via `R_LIBS_USER`) which filters it out of the list, and then the user tries to install into the bootstrap library. Not ideal. Now we create the library when we configure the driver.
* the logging functions did not properly load the root, so failed in odd ways in real usage.

The third bit might change over time, but allows us to create a development set of bootstrap packages that we can use to make sure things work in practice. These can be selected by setting the (undocumented) global option `hipercow.development`. I'll add docs on this process to the administration vignette, once it's written

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ roxygen:
 	${RSCRIPT} -e "devtools::document()"
 
 install:
-	R CMD INSTALL .
+	R CMD INSTALL . drivers/windows
 
 build:
 	R CMD build .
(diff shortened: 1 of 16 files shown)


richfitz (commented):
As discussed, firing this one right through but we can discuss after you get back @weshinsley

PR #34, opened 2023-12-20
Rename authentication function for consistency

richfitz:
* export windows_authenticate() which collects and stores credentials (we're a bit sloppy here actually about what this should be called)
* rename some dide_* functions so that they're the same in hermod and hermod.windows, otherwise it's hard to think about
* make sure the function you are asked to call if not authenticated is the correct one


weshinsley (approved):
OK - that makes sense/makes things clearer!

PR #35, opened 2023-12-20
Add verbose running of tasks

richfitz:
This is needed in order to make the logging do anything interesting at all.


weshinsley (approved):
Looks really nice!

PR #36, opened 2023-12-21
Use logwatch for logs

richfitz:


R/task.R
```diff
@@ -387,6 +399,42 @@ task_log_value <- function(id, root = NULL) {
...
 
 
+##' @rdname task_log
+##'
+##' @inheritParams task_wait
+##'
+##' @inheritParams logwatch::logwatch
+##'
+##' @export
+task_log_watch <- function(id, poll = 1, skip = 0, timeout = Inf,
```
weshinsley, line 409: So skip is not yet supported, but will be...

drivers/windows/DESCRIPTION
```diff
@@ -34,4 +35,5 @@ Suggests:
     withr
 Config/testthat/edition: 3
 Remotes:
-    mrc-ide/conan2@main
+    mrc-ide/conan2@mrc-4834,
```
weshinsley, line 38: Update before merge

DESCRIPTION
```diff
@@ -23,9 +23,11 @@ Imports:
...
     callr,
     conan2,
+    logwatch,
     mockery,
     ps,
     testthat (>= 3.0.0)
 Config/testthat/edition: 3
 Remotes:
-    mrc-ide/conan2@main
+    mrc-ide/conan2@mrc-4834,
```
weshinsley, line 32: Update before merge

weshinsley (approved):
All looks good to me - read through the new logwatch package to see where that code from conan2 went - looks good to have that factored out to be shared...

R/task.R
richfitz: yeah, I think this will be useful when moving things around over ssh potentially

PR #39, opened 2023-12-22
Basic bulk submission support

richfitz:
This PR allows submission of a group of tasks like:

```r
data <- data.frame(a = 1:3, b = c("x", "y", "z"))
task_create_bulk_expr(f(a, b), data)
```

which will end up creating three tasks

```r
f(1, "x")
f(2, "y")
f(3, "z")
```

We do this by binding each row of the df into a local variable that we make available when the task runs.  

I've got the beginnings of rlang's metaprogramming support working here so that the user can use `!!` to pull a value in from the parent environment where things are ambiguous, e.g.

```r
a <- 10
task_create_bulk_expr(f(!!a, b), data)
```

will submit all three tasks with `a` as 10 rather than the value from the df!

We'll certainly tweak this as we go on, but this makes a reasonable starting point I think


weshinsley (approved):
Ok - I had to stare and poke at lines 42-56 a bit, especially about the `!!`, and think I see how it works just about enough!

PR #40, opened 2023-12-22
Report on current configuration

richfitz:
We'll improve this over time, but this gives us something comparable with the previous version at least.


weshinsley (approved):
Looks great! I've added a fix for a typo (hipercow_get_configuration -> hypercow_configuration) in R/configure.R, and re-documented.

PR #41, opened 2023-12-22
Drop unused hipercow.json

richfitz:
We never use this, never add it it, never read it. It's just confusing and we might end up depending on it...


weshinsley (approved):
All looks good!

PR #42, opened 2023-12-22
Fetch "outer" logs

richfitz:
~Merge after #36 as it contains those commits. As in that PR we use a new version of conan2 for the tests (https://github.com/mrc-ide/hipercow/pull/36).~

We've always had access to the "dide log", which is the log from the HPC software itself before we hand off to hipercow.  This PR adds accessing that to the log functions (task_log_value and task_log_show) as a boolean argument `outer`. I don't love the name and we can easily change it, but I do think that putting it in the general task functions is probably preferable to another function.

I've dragged this change through the driver, and added a sample response that checks that accessing a nonexistent task will return the expected value.


weshinsley (approved):
All looking good - tested with a real windows cluster job!

PR #43, opened 2024-01-02
Allow retrying tasks

richfitz:
Merge after #39, contains those commits

Basic support for allowing retrying tasks. When a task is retried we just point back to the original task. Retried tasks always exist in a chain (A -> B -> C) with no forks; so after B has been created then retrying A or B creates C with B as an immediate parent and A as the base task.  This is the same logic as in rrq though implemented quite differently.

I've not put the information about chasing around pointers into the tasks directory because it will result in a lot of reading, and because it will interfere with later on "compacting" completed tasks. Basically; retrying a task does not modify the task but it does set up some level of redirection around identifiers.

Once retried the default behaviour of all functions (getting status, logs, etc) is to follow through to the latest in the chain. This means that users don't need to update things too much if they have a group of tasks and retry some of them.


R/task.R
```diff
@@ -497,7 +513,8 @@ task_log_watch <- function(id, poll = 1, skip = 0, timeout = Inf,
 }
 
 
-task_log_fetch <- function(id, outer, root) {
+task_log_fetch <- function(id, follow, outer, root) {
+  id <- follow_retry_map(id, root)
```
weshinsley, line 517:
```suggestion
  if (follow) {
    id <- follow_retry_map(id, root)
  }
```

weshinsley: Otherwise we get the log for the same id of retried jobs even if follow = FALSE

R/task.R
```diff
@@ -546,10 +563,11 @@ final_status_to_logical <- function(status) {
 ##'   `FALSE` otherwise.
 ##'
 ##' @export
-task_wait <- function(id, timeout = Inf, poll = 1, progress = NULL,
-                      root = NULL) {
+task_wait <- function(id, follow = TRUE, timeout = Inf, poll = 1,
+                      progress = NULL, root = NULL) {
   root <- hipercow_root(root)
-  status <- task_status(id, root = root)
+  id <- follow_retry_map(id, root)
```
weshinsley, line 569: I am getting confused, but I think this one should have an if (follow) round it too - suppose...

job id1 fails.
the retry, job id2, runs for a long time, surprisingly, so we wait for it.

In that case... 
task_wait(id2, follow = FALSE) should wait.
task_wait(id2, follow = TRUE) should wait.
task_wait(id1, follow = TRUE) should wait, because we realise id2 is the new id1
task_wait(id1, follow = FALSE) should probably not wait, but I think above it will.

But this is quite hard to test/replicate!

weshinsley: I suppose there's a question on why one would particularly want to do this forcing follow = FALSE

R/task.R
```diff
@@ -468,9 +483,10 @@ task_log_value <- function(id, outer = FALSE, root = NULL) {
 ##' @inheritParams logwatch::logwatch
 ##'
 ##' @export
-task_log_watch <- function(id, poll = 1, skip = 0, timeout = Inf,
+task_log_watch <- function(id, follow = TRUE, poll = 1, skip = 0, timeout = Inf,
                            progress = NULL, root = NULL) {
   root <- hipercow_root(root)
+  id <- follow_retry_map(id, root)
```
weshinsley, line 489: See below - I *think* if (follow) { ... } - but why you would call this on a dead job with follow = FALSE...

R/task.R
```diff
@@ -388,12 +402,13 @@ task_get_driver <- function(id, root = NULL) {
 ##'
 ##' @return The value of the queued expression
 ##' @export
-task_result <- function(id, root = NULL) {
+task_result <- function(id, follow = TRUE, root = NULL) {
   root <- hipercow_root(root)
+  id <- follow_retry_map(id, root)
```
weshinsley, line 407: if (follow) ?

weshinsley (changes requested):
I needed to stare at this one a bit - so my understanding is...

1. We have a three-column csv file keeping id, parent, base.

2. When we retry a job, the new job is genuinely new, so it has a new id. But we keep a link in the CSV file to the most recent prior retry, which is "parent", and the original job id which we started retrying from in the first place (base)

3. If you have a string of failures of the same job (id1, id2, id3, id4), you could retry *any* of these ids, and you'll get a new job id5, with id4 as its parent, and id1 as its base. This felt a bit confusing at first - although because these are retries of the same job, it doesn't really matter which one is the parent, as they're all the same.

4. task_result always returns the most recent job result no matter which id1..4 it is called with. The results of the retry have the appearance of overwriting the original failed job - almost as if you'd rerun it in place. 

5. task_log similarly returns the log of the most recent retry, no matter which id in the chain you call it with.

OR RATHER - 4 and 5 are true if follow is left to the default TRUE. If follow is manhandled to be FALSE, then I think we treat the ids just as ids without referring to the retry map... 

A few places in the PR where I'm not quite sure if that works, or whether there need to be more catches for if follow is overridden.

richfitz:
Spot on on all that missing logic; all added now and tests added that were failing in the old version

weshinsley (approved):
Excellent!

PR #46, opened 2024-01-03
Add windows username to api and configuration

richfitz:
~Merge after #40, contains those commits~

This PR adds a helper function `windows_username()` which returns the username, and then uses that function to add the username into the reported configuration from `hipercow_configuration()`


weshinsley (approved):
Looks good - additionally tested it by windows_authenticate() with both my "normal" and "admin" usernames, and in each case, correct user name was reported. (Not many people will have two valid windows usernames)

PR #47, opened 2024-01-03
Add windows check helpers

richfitz:
These form the start of our eventual "sitrep" functionality, as well as the "hello" function.  This bit just tests "global" okness.


drivers/windows/R/check.R
```diff
@@ -0,0 +1,63 @@
...
+}
+
+
+windows_check_connection <- function(timeout = 1) {
+  result <- tryCatch({
+    httr::HEAD("https://vault.dide.ic.ac.uk:8200", httr::timeout(timeout))
+  }, error = identity)
+  if (inherits(result, "error")) {
+    cli::cli_alert_danger("Failed to make connection with private network")
+    cli::cli_alert_info("Please check that you have ZScaler enabled")
```
weshinsley, line 57: We could say here...
* Check you have working internet access (we could check for this by pinging 8.8.8.8 or something)
* If you are connecting from outside DIDE, check you have ZScaler enabled

A little faffy - perhaps future PRs. What's here already is the likeliest scenario...

weshinsley (approved):
All working great!

A (slightly) less windows-specific one we could consider for another, is disk-space on wherever your hipercow root sits. For another PR maybe.

richfitz:
Disk space is a great idea; we should work out how to do that portably.

PR #48, opened 2024-01-03
Add hello world helper

richfitz:
This formalises the little hello world thing we do. Once we get the failure diagnosis here too we can use that to help guide people to the best fix.


weshinsley (approved):
All working well!

PR #50, opened 2024-01-03
Add script to check version numbers

richfitz:
Having a battle working out what version I have installed, so time to do this at last. From this point all merges need to increase version number.


weshinsley (approved):
Copied from other working repo

Issue #8, opened 2023-11-29
Overall plan


richfitz:
# package structure

Having a mega package is not ideal, because:

* we're going to have to do a lot of internal namespacing (e.g., `windows_valid_clusters()` vs `valid_clusters()`) and we expect that there won't be much code shared between the different backends/drivers
* the code that we end up running on the cluster will have a ton of extra dependencies (e.g., for the http clients, ssh clients, jsonlite, xml2, etc). This is not ideal as it makes the bootstrap harder to support and upgrades more error prone as they are subject to windows file locking rules
* if we do things in one package, we don't know that our abstraction is ok, and it's hard to rewrite/add a driver

Having a split package is not ideal, because:

* we end up with tight coupling of component packages, which makes it not obvious where to add code, and where to document things - this might be somewhat solvable by using a monorepo though.
* the users have to think about what they are installing and using too much - ideally we have them only interacting with a single package.
* we need to do some complex machinery for registering different "backends"

OTOH, there *is* a complexity for having some sort of support for different backends, so some of this is unavoidable, and it's just a case of working out the most sustainable place to hide it.

In order to have the users interact only with `hermod` we might develop:

* `hermod.core` - the bit that runs on the cluster - create a root, create a generic task, run a task etc.
* `hermod.windows` - the windows driver
* `hermod.linux` - the linux driver
* `hermod` - a fairly lightweight package that imports the others and contains the docs

Lay these out in the repo as

* `core/`
* `windows/`
* `linux/`
* `hermod/` - build the pkgdown from here alone

everything installs from universe, users get all drivers all the time.

# Example interactions

Consider just the simplest use case here - equivalent to the old `obj$enqueue(sessionInfo())` "hello world" that we have used in the past, and different user interactions.

## Windows cluster user

Collect credentials, this is done only once per DIDE password change:

```r
hermod::hermod_dide_credentials()
```

* prompts for keychain password, DIDE username, DIDE password
* validates the password by logging in
* saves the username and password in the keychain

Initialise a repo somewhere, telling it we'll be using windows as the driver

```r
hermod::hermod_init(".", driver = "windows")
```

Alternatively we might have done

```
hermod::hermod_init(".")
hermod::hermod_configure("windows")
```

which would be easier to pass through different arguments.

On the old tools the user would do:
```
share <- didehpc::path_mapping("malaria", "M:", "//fi--didef3.dide.ic.ac.uk/malaria", "M:")
[...]

PR #53, opened 2024-01-05
Allow querying conan installs from driver

richfitz:
This PR depends on https://github.com/mrc-ide/conan2/pull/8 and the branch pointer (633cf37) should be updated before merge.

Use the new functionality in https://github.com/mrc-ide/conan2/pull/7 and https://github.com/mrc-ide/conan2/pull/8 to improve how we interact with provisioning from a the driver.  With this we gain three new functions:

* `hipercow_provision_list` which lists previous installations
* `hipercow_provision_check` which does the same but also checks if we're current or not
* `hipercow_provision_compare` which compares installations (the default behaviour being the last installation done)


tests/testthat/helper-hipercow.R
```diff
@@ -95,26 +97,48 @@ elsewhere_cancel <- function(id, config, path_root) {
...
+    hash <- NULL
+  } else {
+    hash <- rlang::inject(conan2::conan_configure(
+      !!!args,
+      path = path_root,
+      path_lib = file.path("hipercow", "lib"),
+      path_bootstrap = .libPaths()[[1]]))$hash
+  }
+  path_lib <- file.path(config$path, "hipertcow", "lib")
+  conan2::conan_is_current(path_lib, conan_config)
```
weshinsley, line 132: Do elsewhere_provision_list and elsewhere_provision_compare ever get called? 

Typo on line 131, (and 137 below) on hiperTcow
And 132 I think should be `conan2::conan_list(path_lib)`

weshinsley (changes requested):
Just one query with the elsewhere provision functions...

weshinsley (approved):
Splendid

PR #60, opened 2024-01-10
Basic bundle CRUD operations

richfitz:
Nothing makes a bundle yet, but this is easy to add to the bulk function - we might want to change semantics around names etc though perhaps so I've not done that yet.

Nothing uses them; there will eventually be one for task_result, task_status, task_wait, task_info (possibly), task_log_value (possibly). Some of these will have interesting semantics where they might do a reduction so I've not started on that yet.

This just does create/read/update/delete operations plus a little print method. The validation for ids is optional because we won't want to validate a bundle from a bulk operation because we already know it.  It's a fairly long changeset but is mostly pretty mechanical.


DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: hipercow
 Title: High Performance Computing
-Version: 0.2.9
+Version: 0.2.10
```
weshinsley, line 3:
```suggestion
Version: 0.2.12
```

R/bundle.R
```diff
@@ -0,0 +1,133 @@
...
+##' that relate together in some way, and we provide some helper
+##' functions for working with them that save you writing lots of
+##' loops.  Each bundle has a name, which will be randomly generated
+##' if you don't provide one, and a set of task ids.
+##'
+##' @title Create task bundle
+##'
+##' @param ids A character vector of task ids
+##'
+##' @param validate Boolean, indicating if we should check that the
```
weshinsley, line 11: I think elsewhere we say Logical - couple of exceptions on task_eval:19 and windows.R:31

R/bundle.R
```diff
@@ -0,0 +1,133 @@
...
+##'   task ids exist.  We always check that the task ids are
+##'   plausible.
+##'
+##' @param name A string, the name for the bundle.  If not given, then
+##'   a random nmae is generated.
+##'
+##' @param overwrite Boolean, indicating that we should overwrite any
```
weshinsley, line 18: Logical?

drivers/windows/DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: hipercow.windows
 Title: DIDE HPC Support for Windows
-Version: 0.2.9
+Version: 0.2.10
```
weshinsley, line 3:
```suggestion
Version: 0.2.12
```

weshinsley (changes requested):
Just the version number - and a query about Boolean vs Logical.
All looks good!

R/bundle.R
richfitz: sure, I'll make that consistent

richfitz: updated throughout now - I think the idea was that as string was a scalar character, then boolean might be a scalar logical? but that is probably a relationship only in my head

weshinsley (approved):

PR #61, opened 2024-01-10
Add creation of task_resource

weshinsley:
So far, from a hipercow root, you can...

```
resources <- task_resources()
```
and you'll get a list of class `hipercow_resource` with defaults of 1 core, non-exclusive provided by hypercow, and additionally queue = "AllNodes" provided by the windows driver. All the discussed params can be set - validation is not complete yet, but I've done some examples of the formatting validation in hypercow, and the cores/memory limits specific to the windows driver. I'll return to finish that with a quite boring PR later.

Not quite clear yet where to do some of the reformatting/standardising of the various accepted inputs. We might end up with some repeated code in each driver dealing with whether memory is in Gb or Tb for instance. But it's not masses of code...


tests/testthat/test-resource.R
```diff
@@ -0,0 +1,48 @@
...
+  expect_error(validate_cores(-Inf))
+  expect_silent(validate_cores(1))
+  expect_error(validate_cores(-1))
+  expect_error(validate_cores("potato"))
+  expect_error(validate_cores(NULL))
+  expect_error(validate_cores(NA))
+  expect_error(validate_cores(mtcars))
+  expect_error(validate_cores(c(1, 2, 3)))
+
+  expect_error(validate_runtime(c(1, 2, 3)))
```
richfitz, line 12: suggest breaking these into different test_that blocks at exactly the point you have newlines

R/resource.R
```diff
@@ -0,0 +1,160 @@
...
+##'   allow other queuing tasks to jump over them, by setting the priority to
+##'   to `low`; otherwise, the default is `normal`. These are the only
+##'   acceptable values.
+##'
+##' @param queue Specify a particular queue to submit your tasks to.
+##'   This is in development as we decide over time what
+##'   queues we best need for DIDE's common workflows. See the Details
+##'   for more information
+##'
+##' @return A hipercow_resource list, which can be provided when you submit
```
richfitz, line 59: Given this returns a an object of class `hipercow_resource`, it really should be called `hipercow_resource`, not `task_resource`
```suggestion
##' @return A `hipercow_resource` list, which can be provided when you submit
```

R/resource.R
```diff
@@ -0,0 +1,160 @@
...
+task_resources <- function(cores = 1L,
+                           exclusive = FALSE,
+                           runtime = NULL,
+                           hold_until = NULL,
+                           memory_per_node = NULL,
+                           memory_per_process = NULL,
+                           requested_nodes = NULL,
+                           priority = NULL,
+                           queue = NULL,
+                           driver = NULL,
```
richfitz, line 75: I don't know that we want either driver or root here atm, really.  It might be worth considerin the flow here and where we want the validation to happen.  Two options are:

```
r <- task_resources(..., root = .) # validation here
task_create_expr(..., resources = r) # no validation here
```

and 

```
r <- task_resources(...) # no validation here
task_create_expr(..., resources = r, root = .) # validation here
```

Doing it where you do it at the moment requires that we store the driver information and possibly the root to ensure that when we do do the task creation that the same driver and root were used, which feels like it's going to be annoying, so I suggest dropping both args now.

If the motivation is just that the user can check that the resources would be ok, then add a free function `hipercow_resources_validate()` that would do this

R/resource.R
```diff
@@ -0,0 +1,160 @@
...
+##' Coming Soon.
+##'
+##' @title Task Resources
+##'
+##' @param cores The number of cores your task requires. This is 1 by
+##'   default. An error is reported if you request more cores than
+##'   is available on the largest cluster node. Setting to `Inf` will
+##'   request any entire node, however many cores it has.
+##'
+##' @param exclusive Set this to `TRUE` to ensure no other tasks will be
```
richfitz, line 20: Add something about the intent? I know we had a couple of different use cases here.

> This might be useful in the case where your task will use so much RAM that we don't want anything else on the same node, or where ...

R/resource.R
```diff
@@ -0,0 +1,160 @@
...
+##'   concurrently run on the node while it runs your task.
+##'
+##' @param runtime Set this to specify a time limit for running your job.
```
richfitz, line 23: Rename to make it clear it's a limit? Add examples for the `h:m` etc versions?

R/resource.R
```diff
@@ -0,0 +1,160 @@
...
+##'   Acceptable formats are either an integer number of minutes, or
+##'   strings in the form `"h:m"`, or `"d:h:m"` to specify
+##'   days, hours and minutes.
+##'
+##' @param hold_until Specify your task should wait in the queue until
+##'   a certain time, or for a certain period. For the former, this can
+##'   be a POSIXt, a Date, the special strings "tonight" (7pm), "midnight",
```
richfitz, line 30: If it's a date, what time on that date do we start?

richfitz, line 30:
```suggestion
##'   be a [POSIXt], a [Date], the special strings "tonight" (7pm), "midnight",
```

R/resource.R
```diff
@@ -0,0 +1,160 @@
...
+##'   or "weekend" (midnight Saturday morning). To delay for a period,
+##'   this can be an integer number of minutes, or a string in the form
+##'   `"h:m"` for hours and minutes.
```
richfitz, line 33:
```suggestion
##'   `"h:m"` for hours and minutes, as for `runtime`.
```

(or `runtime_limit` if you choose to rename that)

R/resource.R
```diff
@@ -0,0 +1,160 @@
...
+##'
+##' @param memory_per_node Specify your task can only run on a node
+##'   with at least the specified memory. This is an integer,
+##'   measured in gigabytes, or strings such as `"64Gb"` or `"1Tb"`.
+##'
+##' @param memory_per_process If you can provide an estimate of how
+##'   much RAM your task requires, then the cluster can ensure the
+##'   total memory required by running multiple tasks on a node
+##'   does not exceed how much memory the node has. Specify this as
+##'   an integer number of gigabytes, or characters such as `"10Gb"`
```
richfitz, line 43: Is this per core? I think not, right? It might be worth mentioning this. So: 

> If you request `memory_per_process = 64` and `cores = 32` each core will have (effectively) 2GB of RAM. If you are doing OpenMP parallelism this will likely be fine, but if you are parallelising with a higher-level R framework like `future` you may hit memory issues sooner than you expect

It might also be worth mentioning the intent here; this is used by the scheduler to try and keep other tasks away from yours, but only works if everyone uses it, and by the time you're here you might need to also use `exclusive = TRUE`?

R/resource.R
```diff
@@ -0,0 +1,160 @@
...
+##'
+##' @param requested_nodes If you have been in touch with us or DIDE
+##'   IT, and you need to run your task on a selection of named compute
+##'   nodes, then specify this here as a vector of strings for the node names.
+##'
+##' @param priority If the tasks you are launching are low priority, you can
+##'   allow other queuing tasks to jump over them, by setting the priority to
+##'   to `low`; otherwise, the default is `normal`. These are the only
+##'   acceptable values.
```
richfitz, line 52: perhaps we should (in another PR) rickroll people who try `priority = "high"` :smiling_imp:

R/resource.R
```diff
@@ -0,0 +1,160 @@
...
+##'
+##' @param queue Specify a particular queue to submit your tasks to.
+##'   This is in development as we decide over time what
+##'   queues we best need for DIDE's common workflows. See the Details
+##'   for more information
```
richfitz, line 57:
```suggestion
##'   for more information, and the queues available on each cluster.
```

R/resource.R
```diff
@@ -0,0 +1,160 @@
+##' Specify what resources a task requires to run.
+##'
+##' # Windows cluster
+##'
+##' The windows cluster currently is all 32-core nodes, with RAM
+##' mainly 512Gb, and a few 384Gb.
```
richfitz, line 6: You promised to mention available queues here. You might also translate these numbers here into options:

> This means
> 
> * `cores` can take any integer value from 1 to 32 inclusive
> * `memory_per_node` and `memory_per_process` can range up to `512Gb`, but past `384Gb` you may start limiting the nodes you can use

R/resource.R
```diff
@@ -0,0 +1,160 @@
...
+  assert_scalar_logical(exclusive)
+  is.null(runtime) || validate_runtime(runtime)
+  is.null(hold_until) || validate_hold_until(hold_until)
+  is.null(memory_per_node) || validate_memory(memory_per_node)
+  is.null(memory_per_process) || validate_memory(memory_per_process)
+  is.null(requested_nodes) || validate_nodes(requested_nodes)
+  is.null(priority) || validate_priority(priority)
+  is.null(queue) || validate_queue(queue)
+
+  res <- as.list(environment())
```
richfitz, line 94: Please don't do this, you've scooped up `root` and `dat` inadvertently`; see the suggestion at `runtime` which should help with this

R/resource.R
```diff
@@ -0,0 +1,160 @@
...
+  root <- hipercow_root(root)
+  dat <- hipercow_driver_prepare(driver, root, rlang::current_env())
+
+  validate_cores(cores)
+  if (rlang::is_integerish(cores)) {
+    cores <- as.integer(cores)
+  }
+
+  assert_scalar_logical(exclusive)
+  is.null(runtime) || validate_runtime(runtime)
```
richfitz, line 86:
```suggestion
  ret <- list(
    runtime = validate_runtime(runtime),
    ...) # etc
```

then in `validate_runtime` include the null bit:

```
validate_runtime <- function(runtime) {
  if (is.null(runtime)) {
    return(NULL) # happy path, early exit, equivalent to what you had before
  }
  assert_scalar(runtime) # can also pass `call` through here for nice traces
  if (is.integer(runtime)) {
    # check positive etc
  } else if (is.character(runtime)) {
    stop("not yet implemented")
  } else {
    cli::cli_abort("Invalid value for '{runtime}'; expected a NULL, integer or string")
  }
  runtime
}
```

and similarly for the rest of these functions

R/resource.R
```diff
@@ -0,0 +1,160 @@
...
+    mem <- paste0(mem, "Gb")
+  }
+  gb <- grepl("^(\\d+)Gb$", mem)
+  tb <- grepl("^(\\d+)Tb$", mem)
+  if ((!gb) && (!tb)) {
+    cli::cli_abort(c(
+      "Couldn't intepret memory format from `{mem}`.",
+      i = "Examples: 10Gb or 1Tb"))
+  }
+  TRUE
```
richfitz, line 141: returning an integer here, as suggested in the longer comment above, will prevent this logic being duplicated everywhere

richfitz (changes requested):

NAMESPACE
```diff
@@ -26,6 +26,7 @@ export(task_info)
 export(task_log_show)
 export(task_log_value)
 export(task_log_watch)
+export(task_resources)
```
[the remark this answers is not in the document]
weshinsley: I started with hipercow_resources, but steered away as I felt the hipercow_ functions all apply to the hipercow root - (hipercow_init, and the various hipercow_proovision calls. hipercow_hello a bit of an exception!) - so they affect all tasks run from with that hipercow root. 

The resources felt more task-scoped, with a different (perhaps new) resource list created for each bunch of tasks that gets submitted-  so something like task_create_expr(expr, task_resources) felt more natural.

I also toyed with just `resources` but thought it wasn't specific enough - the disk-space available on the networks share is also a resource of a kind and I wondered whether we'd ever want to talk about local resources, or network resources, or... 

cluster_resources feels good enough if task_resources isn't.

R/configure.R
```diff
@@ -90,10 +90,13 @@ hipercow_configure <- function(driver, ..., root = NULL) {
 ##' @param provision_compare Test if a library is current.  It is
 ##'   expected that this will call `conan2::conan_compare`
 ##'
+##' @param task_resources Build a valid task_resources list, which specifies
```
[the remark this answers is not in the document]
weshinsley: My feeling was to try and not have to say "validate" out loud - you just end up with something that is valid - or you don't - when you try to make it. Perhaps cluster_resources is the one? (I am not sure whether resource(s) should be plural)

drivers/windows/R/resource.R
```diff
@@ -0,0 +1,54 @@
...
+validate_queue <- function(queue) {
+  if (!queue %in% c("AllNodes", "Training")) {
+    cli::cli_abort(c(
+      "Queue {queue} is unknown.",
+      i = "Please use AllNodes, or Training"))
+  }
+  TRUE
+}
+
+interpret_memory <- function(mem) {
```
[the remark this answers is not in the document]
weshinsley: Yeah - this is something I'd like to do so we don't have to keep doing the same translations in each driver we write. What paused me was whether we'd like to say "what resources did I run that task with?" and have them exactly as the user said. Eg, they said "hold until tonight, and use 1Tb of RAM" - which we might internally translate to hold until January 10th, 12am, and run on nodes with 1000Gb - would we want to report the original words the user wrote to request the resource.

I guess if we want to return this at all, we could store both in some way, but still do the translations at hipercow level and also have an intermediate representation we pass to any drivers.

R/resource.R
weshinsley: Oh yes... that seemed convenient at the time - I wrote that before I used root and dat...

R/resource.R
weshinsley: Yeah - I started doing this - it was feeling quite sprawly, but is the right way...

R/resource.R
weshinsley: Oh, that should be a DateTime of some sort.

R/resource.R
weshinsley: I wasn't sure otherwise how to find the driver to call the driver-specific resource-list function to fill in the necessary missing defaults / validate against specific cores/ram. But yes, that will mean you could get a resources object and use it invalidly elsewhere...

hmm

R/resource.R
```diff
@@ -0,0 +1,328 @@
...
+}
+
+validate_hold_until <- function(hold_until) {
+  if (is.null(hold_until)) {
+    return(list(original = NULL, computed = NULL))
+  }
+  
+  assert_scalar(hold_until)
+  
+  if ("POSIXt" %in% class(hold_until)) {
```
weshinsley, line 152:
```suggestion
  if (inherits(hold_until, "POSIXt")) {
```

R/resource.R
```diff
@@ -0,0 +1,328 @@
...
+    if (hold_until < Sys.time()) {
+      cli::cli_abort("hold_until time {hold_until} is in the past.")
+    }
+    return(list(original = hold_until, computed = hold_until))
+  }
+  
+  if (class(hold_until) == "Date") {
```
weshinsley, line 159:
```suggestion
  if (is.Date(hold_until)) {
```

R/resource.R
```diff
@@ -0,0 +1,328 @@
...
+
+validate_memory <- function(mem) {
+  if (is.null(mem)) {
+    return(list(original = NULL, computed = NULL))
+  }
+
+  assert_scalar(mem)
+  orig <- mem
+  
+  # If a characterised integer on its own: ("1")
```
weshinsley, line 195: grepl("^[0-9]+$")

_pkgdown.yml
```diff
@@ -50,6 +50,11 @@ reference:
       - task_create_expr
       - task_create_script
       - task_create_bulk_expr
+      
+  - subtitle: Cluster resources 
+    contents:
+      - hipercow_resources
+      - hipercow_validate_resources
```
weshinsley, line 57:
```suggestion
      - hipercow_resources_validate
```

drivers/windows/R/resource.R
```diff
@@ -0,0 +1,13 @@
...
+
+windows_cluster_info <- function(res) {
+  info <- list(
+    max_cores = 32,
+    max_ram = 512,
+    queues = c("AllNodes", "Training"),
+    nodes = sprintf("wpia-%003d", (1:70)[-c(41, 42, 49, 50)])
+  )
+  
+  class(info) <- "windows_cluster_info"
```
weshinsley, line 11: drop?

tests/testthat/test-resource.R
```diff
@@ -0,0 +1,130 @@
...
+  resource_test(validate_queue, NULL)
+  expect_error(validate_queue(NA))
+  expect_error(validate_queue(c("a", "b")))
+  resource_test(validate_queue, "Q")
+})
+
+
+test_that("Can get a hipercow_resource", {
+  res <- hipercow_resources()
+  expect_true("hipercow_resource" %in% class(res))
```
weshinsley, line 86: expect_s3_class

R/resource.R
```diff
@@ -0,0 +1,328 @@
...
+##'   
+##' @export
+##' 
+hipercow_resources_validate <- function(resources,
+                                        driver = NULL,
+                                        root = NULL) {
+  
+  cluster_info <- hipercow_cluster_info(driver, root)
+  
+  cores <- resources$cores$computed
```
weshinsley, line 282: split into small functions

drivers/windows/R/resource.R
```diff
@@ -0,0 +1,13 @@
+## windows-specific cluster functions, called from hipercow.
+
+windows_cluster_info <- function(res) {
```
weshinsley, line 3:
```suggestion
windows_cluster_info <- function(config, path_root) {
```

then switch on `config$cluster`, erroring if it's not known. Ignore the `path_root ` arg

R/resource.R
```diff
@@ -0,0 +1,328 @@
...
+  }
+  
+  assert_scalar_character(queue)
+  list(original = queue, computed = trimws(queue))
+}
+
+hipercow_cluster_info <- function(driver = NULL, root = NULL) {
+  root <- hipercow_root(root)
+  dat <- hipercow_driver_prepare(driver, root, rlang::current_env())
+  dat$driver$cluster_info()
```
weshinsley, line 258:
```suggestion
  dat$driver$cluster_info(dat$config, root$path$root)
```

R/resource.R
```diff
@@ -0,0 +1,328 @@
...
+##' @export
+##' 
+hipercow_resources_validate <- function(resources,
+                                        driver = NULL,
+                                        root = NULL) {
+  
+  cluster_info <- hipercow_cluster_info(driver, root)
+  
+  cores <- resources$cores$computed
+  if ((cores != Inf) && (cores > cluster_info$max_cores)) {
```
weshinsley, line 283:
```suggestion
  if (cores != Inf && cores > cluster_info$max_cores) {
```

R/util.R
```diff
@@ -237,3 +237,42 @@ time_ago <- function(time, missing = "unknown time ago") {
     paste(format(Sys.time() - time, digits = 2), "ago")
   }
 }
+
+duration_to_minutes <- function(period) {
```
weshinsley, line 241: pass a name here, also work out why fail_msg did not fail. Perhaps this line is not covered?

richfitz (approved):

PR #62, opened 2024-01-11
mrc-4912: Allow waiting on tasks until they start running

richfitz:
This generalises `task_wait` a little by allowing us to wait until a task is at least running.  This has two uses:

* I was keen to use this in a vignette to cancel a running task (rather than just one that is waiting)
* We will want this in rrq perhaps to want until tasks are running (and not until they terminate)


weshinsley (approved):
Nice easy one!

PR #63, opened 2024-01-12
Use hipercow_resources when running tasks

weshinsley:
A hipercow_resource can now be passed into all the task_create functions, and ultimately task_submit - OR - it can be left NULL and a working default will be picked up.

task_retry is a bit interesting - I haven't done anthing special.

Also - I have one mockery I couldn't get working in test-interface - I wanted to check that special_time got called once within task_submit - and I'm sure it does from manual browser()-ing - but I've done something wrong with the mock and it doesn't count the call...


R/resource.R
```diff
@@ -265,8 +265,21 @@ hipercow_resources_validate <- function(resources,
                                         driver = NULL,
                                         root = NULL) {
 
+  # A bit of a hack for now - handle NULL driver (which often happens
```
richfitz, line 268: can you make a ticket with in the "debt and maintenance" epic?

R/resource.R
```diff
@@ -265,8 +265,21 @@ hipercow_resources_validate <- function(resources,
...
+  # "driver must be a scalar" error.
+  
+  if (is.null(driver) || length(driver) > 1) {
+    return(TRUE)
+  }
+  
   cluster_info <- hipercow_cluster_info(driver, root)
 
+  if (is.null(resources$queue$computed)) {
+    resources$queue$computed <- cluster_info$default_queue
```
richfitz, line 280: do we have this removed now from the configuration?

R/task-create.R
```diff
@@ -36,14 +39,18 @@
 ##' @export
 task_create_explicit <- function(expr, export = NULL, envir = .GlobalEnv,
                                  environment = "default", submit = NULL,
-                                 root = NULL) {
+                                 resources = NULL, root = NULL) {
   root <- hipercow_root(root)
+  if (is.null(resources)) {
+    resources <- hipercow_resources()
+  }
+  resources <- hipercow_resources_validate(resources, names(root$config), root)
```
richfitz, line 47: These 4 lines will be everywhere. Add a new helper:

```
as_hipercow_resources <- function(x) {
  if (is.null(resources)) {
    resources <- hipercow_resources()
  }
  hipercow_resources_validate(resources, names(root$config), root)
}
```

in resources.R; there's more good we can do with this later too

R/submit.R
```diff
@@ -10,20 +10,28 @@
...
+task_submit <- function(id, ..., resources = NULL, 
+                        driver = NULL, root = NULL) {
   if (...length() > 0) {
     cli::cli_abort("Additional arguments to 'task_submit' not allowed")
   }
   root <- hipercow_root(root)
-
   dat <- hipercow_driver_prepare(driver, root, environment())
-
+  if (!is.null(resources$hold_until$computed)) {
```
richfitz, line 29: I feel this could be done at validation, unless someone sits on their resources a long time.

R/task-retry.R
```diff
@@ -23,8 +23,16 @@
 ##'
 ##' @export
 ##' @return New identifiers for the retried tasks
-task_retry <- function(id, submit = NULL, root = NULL) {
+task_retry <- function(id, submit = NULL, resources = NULL, root = NULL) {
   root <- hipercow_root(root)
+  
+  # More thinking to do on what resources should be for a retry
```
richfitz, line 29: Can you make a ticket? we should probably chat through this one to find a reasonable solution

R/util.R
```diff
@@ -282,4 +282,57 @@ duration_to_minutes <- function(period, name = "testing") {
...
+}
+
+to_POSIXct <- function(s) {
+  as.POSIXct(s, format = "%Y-%m-%d %H:%M:%S")
+}
+
+
+special_time <- function(name, now = Sys.time()) {
+  break_up <- function(dt) {
+    list(year = as.integer(format(dt, "%Y")),
```
richfitz, line 300: helpful?

```
unclass(as.POSIXlt(Sys.time()))
```

drivers/windows/R/driver.R
```diff
@@ -14,15 +14,15 @@ hipercow_driver_windows <- function() {
...
+windows_submit <- function(id, resources, config, path_root) {
   path_batch <- write_batch_task_run(id, config, path_root)
 
   path_batch_dat <- prepare_path(path_batch, config$shares)
   path_batch_unc <- windows_path_slashes(
     file.path(path_batch_dat$path_remote, path_batch_dat$rel))
 
   client <- get_web_client()
-  dide_id <- client$submit(path_batch_unc, id, config$template)
+  dide_id <- client$submit(path_batch_unc, id, resources)
```
richfitz, line 25: Can you remove template through the configuration of the driver? Or just make a ticket and we can do it later

drivers/windows/R/resource.R
```diff
@@ -6,6 +6,7 @@ windows_cluster_info <- function(config, path_root) {
       max_cores = 32,
       max_ram = 512,
       queues = c("AllNodes", "Training"),
+      default_queue = "AllNodes",
```
richfitz, line 9: or, have the first element of queues be default?

richfitz (changes requested):
quick review to give you something to get on with on this in case I'm off monday

R/resource.R
weshinsley: Done now - along with valid_templates from clusters.R in the driver, which we no longer need

drivers/windows/R/driver.R
weshinsley: Think this is now done

R/util.R
weshinsley: A little bit!

R/task-create.R
weshinsley: Done - as_hipercow_resources(resources, root)

drivers/windows/R/resource.R
weshinsley: Hmm - I think I maybe prefer saying explicitly `default_queue` than `queues[1]` but not strongly so...

R/task-retry.R
weshinsley: Let's use yours - 
https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4941/Resources-workflow-for-retried-tasks

R/resource.R
weshinsley: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4942

R/submit.R
weshinsley: I am in two minds about this. I wonder if people might create a couple of resource objects for their "test_job" and their "real_job" and their "massive_job_that_always_waits_til_midnight", and then have those as kind of ready-made resources to launch their jobs in the same way as last time, without having to think about it.

I am not sure whether this is terribly likely, and might not be totally advisable - once in a while, available resources might change (but then they will need a new package version and will have to rerun their resource things anyway).

 But also, I'm not sure doing the translation during submission causes any real problem does it?

richfitz: the only "problem" caused by doing it at submission is that there's two places to do the validation really.  In particular, the validation here must never fail or it will be quite annoying for the user.  We can see how this goes in practice if you want - I expect this flow will need some adjustment when it comes in contact with users, no matter what we pick.

We might sketch out a flow of validation etc as we complicate this too (environment variables and parallel configuration will all interact here too)

tests/testthat/test-interface.R
```diff
@@ -39,6 +39,37 @@ test_that("can submit a task via a driver", {
...
+
+  mock_special_time <- mockery::mock("midnight2")
+  mockery::stub(task_submit, "special_time", mock_special_time)
+
+  expect_message(
+    id <- withr::with_dir(path_here, task_create_explicit(quote(getwd()),
+                        resources = res)),
+    "Submitted task '[[:xdigit:]]{32}' using 'elsewhere'")
+
+  # TODO HERE - Something is wrong with my mockery - mock_special_time
```
richfitz, line 62: I'll talk you through this in person, it's an unfortunate consequence of how mockery works.

Usually some contortions are required here to make good tests

richfitz (approved):

PR #64, opened 2024-01-12
Implement basic bundle operations

richfitz:
This PR generalises the task operations to work with bundles.  So `task_result` has a `hipercow_bundle_result` and so on.  The only one I've not done is `task_info` which feels different.

* Note that now `task_create_bulk_expr` now returns a bundle, and bundle creation prints the name of the bundle
* I can split this into small PRs, function by function easily enough if this is overwhelming
* I wonder if we should do `bundle_*` for all these functions
* I see a future issue with `hipercow_bundle_delete` once we have `task_delete`.  Perhaps we will use `task_purge` which is more evocative
* All bundle functions have follow as do the task functions
* The status reduction bit turns out to be nice, and somewhat cute even

Differences from the underlying functions:

* `status` gets a `reduce` argument
* `wait` gets a `fail_early` argument
* `retry` gets an `if_status_in` argument


DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: hipercow
 Title: High Performance Computing
-Version: 0.2.16
+Version: 0.2.17
```
weshinsley:
```suggestion
Version: 0.2.18
```

drivers/windows/DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: hipercow.windows
 Title: DIDE HPC Support for Windows
-Version: 0.2.16
+Version: 0.2.17
```
weshinsley:
```suggestion
Version: 0.2.18
```

R/bundle.R
```diff
@@ -120,14 +121,255 @@ hipercow_bundle_delete <- function(name, root = NULL) {
...
+##'
+##' @return A list with each element being the logs for the
+##'   corresponding element in the bundle.
+##'
+##' @export
+hipercow_bundle_log_value <- function(bundle, follow = TRUE, outer = FALSE,
+                                      root = NULL) {
+  root <- hipercow_root(root)
+  bundle <- check_bundle(bundle, root, rlang::current_env())
+  lapply(bundle$ids, task_log_value, outer = outer, root = root)
```
weshinsley, line 279:
```suggestion
  lapply(bundle$ids, task_log_value, follow = follow, outer = outer,
         root = root)
```

R/bundle.R
```diff
@@ -120,14 +121,255 @@ hipercow_bundle_delete <- function(name, root = NULL) {
...
+    if (length(err) > 0) {
+      cli::cli_abort("Invalid value{?s} for 'if_status_in': {squote(err)}")
+    }
+  }
+  status <- hipercow_bundle_status(bundle, follow = TRUE, root = root)
+  i <- status %in% if_status_in
+  if (any(i)) {
+    cli::cli_alert_info("Retrying {sum(i)} / {length(i)} task{?s}")
+    ids <- vcapply(bundle$ids[i], task_retry, submit = FALSE, root = root)
+    task_submit_maybe(ids, submit, root, rlang::current_env())
```
weshinsley, line 326:
```suggestion
    task_submit_maybe(ids, submit, resources = NULL, root, rlang::current_env())
```

weshinsley: (Resources got merged under this one)

weshinsley (changes requested):
Couple of quick things!

PR #68, opened 2024-01-16
Assorted improvements

richfitz:
Everything but the moo


R/hipercow.R
```diff
@@ -1,3 +1,8 @@
+##' HiperCow
+##' @title HiperCow
+##' @return Moo
+##' @export
+##' @keywords internal
 hipercow <- function() {
   nrow <- logo_cols$dim[[1]]
```
weshinsley, line 7: For improved experience in RStudio, could consider doing this somewhere in the installation/creation of logo_cols - although there are still thin white lines between each row, so RStudio is still a bit suboptimal
```
logo_cols$open <- paste0("\033[1m\033[48;5;0m", logo_cols$open)
```
\033[1m does bold
\033[48;5;0m does black background

PR #69, opened 2024-01-16
Allow multiline expressions

richfitz:
This is going to be useful for implementing `rrq`/workers support, and possibly for some advanced users.  The idea is to allow multi-line expressions contained in `{}` so for example

```r
task_create_expr({
  x <- 1
  f(x)
})
```

which will run the entire expression, "returning" the last line (this is the behaviour of `local` already).  We actually supported this already, just not usefully because in this case we should exclude `x` from our list of local variables; that's the meat of this PR.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: hipercow
 Title: High Performance Computing
-Version: 0.2.21
+Version: 0.2.22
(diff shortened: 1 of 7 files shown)


weshinsley (changes requested):
Ok - think I get this - just a quick typo/missing word in the docs

weshinsley (approved):

PR #70, opened 2024-01-17
Basic support for envvars in tasks

richfitz:
This PR allows passing envvars into tasks.  The approach is similar to resources, except that we save these into the core "expr" object.  We might allow some system for overriding these with retries but for now this should be reasonable.

I hit the exact same issue of "I expect exactly one driver configured" that you hit with resources, and


weshinsley (approved):
Really nice!

PR #71, opened 2024-01-17
Consistent error messages for ids and bundle arguments

richfitz:
Just a helper function that will point people in the right direction if they pass a bundle to a task function by accident


weshinsley (approved):
Looks good!

PR #72, opened 2024-01-17
Fix lint

richfitz:
Fixes lint, spelling, and one small NOTE from R CMD check


R/util.R
```diff
@@ -291,12 +291,12 @@ duration_to_minutes <- function(period, name = "testing") {
 
 
 format_datetime <- function(year, month, day, hour, minute, second) {
-  format(to_POSIXct(
+  format(to_posix_ct(
     sprintf("%s-%s-%s %s:%s:%s", year, month, day, hour, minute, second)),
     "%Y-%m-%d %H:%M:%S")
 }
 
-to_POSIXct <- function(s) {
```
weshinsley, line 299: Hmm, not sure of these two since POSIXct is the name of the R type... but ok...!

drivers/windows/R/web.R
```diff
@@ -236,33 +236,33 @@ client_body_submit <- function(path, name, resources, cluster,
               jobs = encode64(path_call),
               dep = encode64(deps),
               hpcfunc = "submit")
-  
+
   if (resources$cores$computed == Inf) {
```
weshinsley, line 240: Elswhere did we switch to `identical` for comparing to Inf? 
We only want positive Inf, so... is.infinite & > 0 is a wordy option...

weshinsley (approved):
All good. Lots of whitespace I left behind - RStudio not handling that for me even though it is clearly ticked in my prefs.

drivers/windows/R/web.R
richfitz: identical was added because the value could be `NULL`

richfitz: for testing against `-Inf` you'd use `x == -Inf`

PR #79, opened 2024-01-22
System for enabling examples, along with many examples

richfitz:
Sorry about the size of this; I can make it much smaller by dropping the examples, but they were helpful in designing the system.

There is some fallout coming from this in other PRs (but not dependent)

* #78 
* #80 

I'm also trying to tidy up coping with multiple drivers as users will be able to get into that situation more easily now.

Once we're happy with this, we can go through the vignettes and port them over to use the example driver I think, which will help with running vignettes online


R/task-create.R
```diff
@@ -1,4 +1,4 @@
-##' Create an explicit task. Explicit tasks are the simplest sort of
+1##' Create an explicit task. Explicit tasks are the simplest sort of
```
weshinsley, line 1:
```suggestion
##' Create an explicit task. Explicit tasks are the simplest sort of
```

R/task-eval.R
```diff
@@ -18,6 +18,18 @@
 ##'
 ##' @return Logical indicating success (`TRUE`) or failure (`FALSE`)
 ##' @export
+##' @examples
+##' cleanup <- hipercow_example_helper(runner = FALSE)
+##' id <- task_create_expr(runif(1), submit = FALSE)
+##' # Status is only 'created', not 'submitted'; as we did not submit
```
weshinsley, line 24:
```suggestion
##' # Status is only 'created', not 'submitted', as we did not submit
```

R/task.R
```diff
@@ -385,6 +450,17 @@ task_wait <- function(id, follow = TRUE, for_start = FALSE,
 ##'   completed, not running, etc.
 ##'
 ##' @export
+##' @examples
+##' cleanup <- hipercow_example_helper()
+##'
+##' ids <- c(task_create_expr(Sys.sleep(2)), task_create_expr(runif(1)))
+##'
+##' # The first task may or not be cancelled (depends on if it was
+##' # started already) but the second one will almos certainly be
```
weshinsley, line 459:
```suggestion
##' # started already) but the second one will almost certainly be
```

R/windows.R
```diff
@@ -63,6 +70,22 @@ windows_check <- function() {
 ##'   Perhaps there are some guidelines for this somewhere?
 ##'
 ##' @export
+##' @examplesIf FALSE
+##'
+##' # Suppose that have mounted your malaria share at "~/net/malaria"
```
weshinsley, line 75:
```suggestion
##' # Suppose you have mounted your malaria share at "~/net/malaria"
```

PR #80, opened 2024-01-23
Allow removing drivers

richfitz:
This is going to be helpful with the examples being added in #79, as the users might accidentally add the example driver and need to remove it.


R/configure.R
```diff
@@ -32,6 +32,8 @@
 ##'
 ##' @param root Hipercow root, usually best `NULL`
 ##'
+##' @seealso [hipercow_unconfigure], which removes a driver
```
weshinsley, line 35:
```suggestion
##' @seealso [hipercow_unconfigure], which removes a driver, and to see the 
##'    currently enabled drivers, use [hipercow_configuration]
```

weshinsley (approved):
All looks good - one small doc suggestion - we could mention hipercow_configuration lets you see what drivers are currently enabled

PR #81, opened 2024-01-24
Better support for multiple drivers

richfitz:


R/resource.R
```diff
@@ -280,21 +289,32 @@ hipercow_cluster_info <- function(driver = NULL, root = NULL) {
...
+}
 
-  if (is.null(driver) || length(driver) > 1) {
-    return(resources)
+
+resources_validate <- function(resources, driver, root) {
+  given_resources <- !is.null(resources)
+  if (given_resources) {
+    assert_is(resources, "hipercow_resource")
+    if (identical(attr(resources, "validated", exact = TRUE), driver)) {
```
weshinsley, line 303:
```suggestion
    if (!is.null(driver) && 
        (identical(attr(resources, "validated", exact = TRUE), driver))) {
```

weshinsley: Otherwise, if attr(resources, "validated") is NULL (ie, it wasn't validated), then this can match when driver is NULL as well, and we return early, rather than erroring because of resources specified but no driver.

weshinsley (changes requested):
One fix

weshinsley (approved):

PR #84, opened 2024-01-25
Avoid accidentally finding the package root

richfitz:
This PR stops me accidentally finding the hipercow *package* root when writing tests, when what I am expecting to find is the root that we are writing tests against.

I've also dropped rprojroot because it's not buying us here, and just pulled in the same code as in orderly/orderly2 that has served well over the years


weshinsley (approved):
Excellent!

PR #85, opened 2024-01-25
Use example driver in vignettes

richfitz:
This PR updates the vignettes so that only the windows vignette now builds on the cluster itself, all the others will use the example driver and automatically rerun.  Most of the content has been preserved, some has been squashed together.

I was unable to get the packages vignette to build online for reasons that I was unwilling to investigate, so have left that for another day. They build fine locally and the vignette now just uses the example driver so it's much faster to develop.


vignettes_src/windows.Rmd
```diff
@@ -0,0 +1,245 @@
...
+source("../vignettes/common.R")
+vignette_root <- new_hipercow_root_path(TRUE)
+set_vignette_root(vignette_root)
+```
+
+```{r, echo = FALSE, results = "asis"}
+add_header()
+```
+
+We support one cluster at present (the "windows" cluster) but will support more in future; in particular we aim to support a Linux cluster within DIDE ("hermod") and the central ICT/RCS cluster.  In future we may support others, but we have immediate plans to.
```
weshinsley, line 20:
```suggestion
We support one cluster at present (the "windows" cluster) but will support more in future; in particular we aim to support a Linux cluster within DIDE ("hermod") and the central ICT/RCS cluster.  In future we may support others, but we have no immediate plans to.
```

vignettes_src/windows.Rmd
```diff
@@ -0,0 +1,245 @@
...
+* **macOS**: The path will likely be below `/Volumes` (but not one that corresponds to a USB stick of course!)
+* **Linux**: The path should start at one of the mount points you have configured in `/etc/fstab` as a network path.
+
+## Project shares and home directories
+
+In general we **strongly recommend** that you use project shares for any serious work, rather than your home directory.  To organise these you should talk to your PI (or if you are a PI, talk to Chris).  The advantages of the project shares is that they are larger (so you will run out of disk space more slowly) and faster than the home shares.  If you launch many tasks at once that use your home share you can get unexpected failures as the disk can't keep up with amount of data being read and written.  Don't use your home directory (generally `Q:` on windows) for anything more intensive than casual experimentation.  We have seen many people with jobs that fail mysteriously on their network home directory when launched in parallel, and this is generally fixed by using a project share.
```
weshinsley, line 108:
```suggestion
In general we **strongly recommend** that you use project shares for any serious work, rather than your home directory.  To organise these you should talk to your PI (or if you are a PI, talk to Chris).  The advantages of the project shares are that they are larger (so you will run out of disk space more slowly) and faster than the home shares.  If you launch many tasks at once that use your home share you can get unexpected failures as the disk can't keep up with amount of data being read and written.  Don't use your home directory (generally `Q:` on windows) for anything more intensive than casual experimentation.  We have seen many people with jobs that fail mysteriously on their network home directory when launched in parallel, and this is generally fixed by using a project share.
```

vignettes_src/windows.Rmd
```diff
@@ -0,0 +1,245 @@
...
+
+In Finder, go to `Go -> Connect to Server...` or press `Command-K`.
+In the address field write the name of the share you want to
+connect to.  Useful ones are
+
+* `smb://fi--san03.dide.ic.ac.uk/homes/<username>` -- your home share
+* `smb://fi--didef3.dide.ic.ac.uk/tmp` -- the temporary share
+
+At some point in the process you should get prompted for your
+username and password, but I can't remember what that looks like.
```
weshinsley, line 136: TO-DO - we should improve that last sentence a bit. 
Maybe also we should avoid talking about the temp drive, or mention that it might be terrible.

weshinsley (approved):

PR #87, opened 2024-01-26
mrc-4951: harmonise errors from resource checks

richfitz:
This PR harmonises and tests errors thrown during resource validation, passing the call through for nicer tracebacks and trying to reflect the argument name back.  The main aim was to update calls to `expect_error()` to add strings that match the error.

Churny but I did manage to simplify some of the time handling, which R makes particularly painful.  This avoids a few roundtrips through strings, and is I think a little easier to reason about though there is quite a bit of fighting terrible classes still.

While doing this I was wondering where the use of "original" is but did not really see anything obvious. Do you remember?

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -313,45 +313,41 @@ duration_to_minutes <- function(period, name = "testing", call = NULL) {
 }
 
 
-format_datetime <- function(year, month, day, hour, minute, second) {
-  format(to_posix_ct(
-    sprintf("%s-%s-%s %s:%s:%s", year, month, day, hour, minute, second)),
-    "%Y-%m-%d %H:%M:%S")
-}
-
-to_posix_ct <- function(s) {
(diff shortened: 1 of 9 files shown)


R/util.R
```diff
@@ -263,91 +263,91 @@ time_ago <- function(time, missing = "unknown time ago") {
...
-  # or if we start with 'm', 'd' or 'h'
+  if (is.numeric(period)) {
+    if (!rlang::is_integerish(period)) {
+      fail_msg("'{name}' is a non-integer number of minutes")
+    }
+    if (period < 0) {
+      fail_msg("'{name}' is a negative number of minutes")
+    }
+    if (period == 0) {
+      fail_msg("{name}' is a zero minutes")
```
weshinsley, line 284:
```suggestion
      fail_msg("{name}' is zero minutes")
```

weshinsley (approved):

PR #88, opened 2024-01-29
Support for sending the hello job to fast queue

richfitz:
This PR organises sending the hello world job through the fast queue.   No bootstrap update required, but we need to ensure that the windows driver and hipercow packages are both installed for it to work.

* new driver function that runs preflight checks and returns resources; this will be pretty general really
* the windows preflight checks now also check that the path is suitable
* resources are now saved as rds into the task directory, we can start using that when doing retries I think

There's one failure on windows only - Wes could you investigate? I think that it might just be that we need to do:

```
if (is_windows()) {
  paths<- windows_path(...)
}
```

in the tests or implementation but I am not really sure what is triggering this

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: hipercow
 Title: High Performance Computing
-Version: 0.2.36
+Version: 0.2.37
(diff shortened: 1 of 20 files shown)


weshinsley (approved):
All looks good

PR #92, opened 2024-01-31
Drop "original" from all computed resources

richfitz:
Seems ok throughout with this change.

Also added a print method for `hipercow_resources` and `hipercow_parallel`; both very simple and using the same little utility function


weshinsley (approved):
That actually tidies things a lot!

PR #94, opened 2024-02-01
Have main thread call hipercow_parallel_set_cores

weshinsley:


R/parallel.R
```diff
@@ -67,29 +67,42 @@ hipercow_parallel_get_cores <- function() {
...
-##' launch using [hipercow_parallel] will have the correct resources, but you
+##' the given number of cores. This is used to help various thread-capable
+##' packages use the correct number of cores. You
 ##' can also call it yourself if you know specifically how many cores you want
 ##' to be available to code that looks up these environment variables.
 ##'
 ##' @title Set various environment variables that report the number of cores
 ##' available for execution.
 ##'
+##' @inheritParams task_eval
```
richfitz, line 78: Need a correct doc entry here for `@envir` ; the one from task_eval is not correct!

Here it's to scope the lifetime of the environment variable changes, perhaps with a hint to the user that they leave it alone or consult `withr::local_envvar` for details

R/parallel.R
```diff
@@ -67,29 +67,42 @@ hipercow_parallel_get_cores <- function() {
...
+##'
 ##' @param cores Number of cores to be used.
 ##'
 ##' @export
-hipercow_parallel_set_cores <- function(cores) {
+hipercow_parallel_set_cores <- function(cores, envir = NULL) {
+  if (is.null(cores)) {
```
richfitz, line 84: I think this should not pass here; this is probably an error, no? So accept here a positive integer only?

R/task-eval.R
```diff
@@ -66,6 +66,11 @@ task_eval <- function(id, envir = .GlobalEnv, verbose = FALSE, root = NULL) {
       data$id <- id
     }
     envvars_apply(data$envvars, top)
+
+    default_cores <- hipercow_parallel_get_cores()
+    if (is.na(default_cores)) default_cores <- NULL
+    hipercow_parallel_set_cores(default_cores, rlang::current_env())
```
richfitz, line 72:
```suggestion
    cores <- hipercow_parallel_get_cores()
    if (!is.na(cores)) {
      hipercow_parallel_set_cores(cores, rlang::current_env())
    }
```

* I don't think this is really *default* cores, so much as cores
* the logic currently in set_cores is better here I think, and avoids churn while passing around different types of missingness

tests/testthat/test-hipercow.R
```diff
@@ -1,3 +1,5 @@
 test_that("hipercow", {
+  mock_play <- mockery::mock(FALSE)
+  mockery::stub(hipercow, "hipercow_speak", mock_play)
   expect_output(hipercow(), "H.+I.+P.+E.+R.+C.+O.+W")
```
richfitz, line 4: Do you want to check that play was called here?

Why are you adding a `FALSE` return value to your mock function? I think you simply want `mock_play <- mockery::mock()` here and in the previous

richfitz (changes requested):

tests/testthat/test-hipercow.R
weshinsley: Oops - copy/paste error on the FALSE. Not very much logic around the plays, but I've added expected call counts.

R/parallel.R
weshinsley: Ok - I think that's better now?

R/parallel.R
weshinsley: I'm getting a little confused by whether cores can be NULL or NA - I think it's probably ok to just take whatever hipercow_parallel_get_cores() returns (which is integer or NA, and we seem to be expecting a potential NA a couple of lines down...)...

See if this looks better now - I think the only place the NULL was coming from, was the new code I wrote, which I've now tamed below...

R/task-eval.R
weshinsley: OK - I think that's better now?

R/task-eval.R
```diff
@@ -66,6 +66,10 @@ task_eval <- function(id, envir = .GlobalEnv, verbose = FALSE, root = NULL) {
       data$id <- id
     }
     envvars_apply(data$envvars, top)
+
+    hipercow_parallel_set_cores(hipercow_parallel_get_cores(),
+                                rlang::current_env())
```
richfitz, line 71:
```suggestion
    cores <- hipercow_parallel_get_cores()
    if (!is.na(cores)) {
      hipercow_parallel_set_cores(cores, rlang::current_env())
    } 
```

R/parallel.R
```diff
@@ -77,19 +77,31 @@ hipercow_parallel_get_cores <- function() {
...
 ##' @param cores Number of cores to be used.
 ##'
+##' @param envir Environment in which the variables will be set to limit their
+##' lifetime. This should not need setting in general, but
+##' see `withr::local_envvar` for example use.
+##'
 ##' @export
-hipercow_parallel_set_cores <- function(cores) {
+hipercow_parallel_set_cores <- function(cores, envir = NULL) {
   prev <- hipercow_parallel_get_cores()
```
richfitz, line 86: Assert that this is a nonmissing scalar integer, etc

richfitz (changes requested):

R/parallel.R
```diff
@@ -77,19 +77,31 @@ hipercow_parallel_get_cores <- function() {
...
   if (!is.na(prev) && (!is.na(cores)) && (cores > prev)) {
```
richfitz, line 87:
```suggestion
  if (!is.na(prev) && cores > prev) {
```

tests/testthat/test-parallel.R
```diff
@@ -26,7 +26,8 @@ test_that("Increasing core count gives a friendly warning", {
   hipercow_parallel_set_cores(1)
   res <- evaluate_promise(hipercow_parallel_set_cores(2))
   expect_true(grepl("(.*)increasing cores alone(.*)", res$messages))
-  expect_silent(hipercow_parallel_set_cores(NA))
+  expect_error(hipercow_parallel_set_cores(NA))
+  expect_error(hipercow_parallel_set_cores(-1))
```
richfitz, line 30: expect_error should always have a string to test the error against

richfitz (approved):

PR #98, opened 2024-02-05
Small doc tweaks

richfitz:
Most of the fixes for Rob's review, I think.

More working hyperlinks, based on how pkgdown does autolinking

Spelling and linting

### DIFF
--- a/vignettes/hipercow.Rmd
+++ b/vignettes/hipercow.Rmd
@@ -239,7 +239,7 @@ task_wait(id)
 task_info(id)
 ```
 
-Both of our parallel tasks are to sleep for 5 seconds. We use `task_info` to report  how long it took for those two runs to execute; if they ran one-by-one, we'd expect around 10 seconds, but we are seeing a much shorter time than that, so our pair of processes are running at the same time.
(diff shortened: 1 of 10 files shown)


vignettes/hipercow.Rmd
```diff
@@ -237,7 +239,7 @@ task_wait(id)
 task_info(id)
 ```
 
-Both of our parallel tasks are to sleep for 5 seconds. We use `task_info` to report  how long it took for those two runs to execute; if they ran one-by-one, we'd expect around 10 seconds, but we are seeing a much shorter time than that, so our pair of processes are running at the same time. 
+Both of our parallel tasks are to sleep for 5 seconds. We use `task_info` to report  how long it took for those two runs to execute; if they ran one-by-one, we'd expect around 10 seconds, but we are seeing a much shorter time than that, so our pair of processes are running at the same time.
```
weshinsley, line 242:
```suggestion
Both of our parallel tasks are to sleep for 5 seconds. We use `task_info()` to report  how long it took for those two runs to execute; if they ran one-by-one, we'd expect around 10 seconds, but we are seeing a much shorter time than that, so our pair of processes are running at the same time.
```

weshinsley (approved):

PR #100, opened 2024-02-06
Validate that bundle and environment names are safe

richfitz:
Getting ahead of the inevitable saving of an environment or bundle with a `/`, `\` or `:` in it, which will fail badly.

Also adding in the idea of an "empty" environment, which comes from part of the rrq PR that I've paused until I get more of the rrq refactor finished


weshinsley (approved):
Looks good!

PR #101, opened 2024-02-07
Shard the path where logs are diverted to in task_run.bat

weshinsley:


drivers/windows/R/check.R
```diff
@@ -109,9 +109,9 @@ windows_check_project <- function(path) {
         "Your working directory is a subdirectory of the project")
       TRUE
     } else {
-      cli::cli_alert_danger(
+      cli::cli_alert_info(
```
richfitz, line 112: Can you undo this change please?

richfitz (changes requested):
Good spot, I really should have tested #96 on the real cluster 🤦

richfitz (approved):

PR #105, opened 2024-02-08
Check cores > 1 if parallel is set

weshinsley:


R/submit.R
```diff
@@ -26,7 +26,7 @@ task_submit <- function(id, ..., resources = NULL,
   }
   root <- hipercow_root(root)
   driver <- hipercow_driver_select(driver, TRUE, root, rlang::current_env())
-  resources <- resources_validate(resources, driver, root)
+  resources <- resources_validate(resources, NULL, driver, root)
```
weshinsley, line 29: I am not sure about this one in the case where task_submit is called directly - (not via one of the task_create calls) - does this ever happen?

weshinsley: I could just name the args driver = driver and root = root, which looks nicer than a naked NULL, but whether task_submit should actually take parallel as an argument...

richfitz: Submit should not need parallel, I think - I think this is related to our nagging concern we have some of the logic of where we save and reference this data from slightly wrong (as discussed with the retry).  For now, `NULL` is fine, name it as an argument though I think.

Another way of doing this that might be nicer is not to do the validation here within resources_validate but do it within the task creation *after* `resources_validate` but before submission?

weshinsley: Yeah - I think this is nicer now?

tests/testthat/test-parallel.R
```diff
@@ -147,9 +147,26 @@ test_that("can print parallel control", {
                all = FALSE, fixed = TRUE)
 })
 
+
+test_that("Can't run parallel with 1 core", {
+  expect_message(
```
richfitz, line 152: Normally here we'd start the example with

```
  path <- withr::local_tempdir()
  init_quietly(path, driver = "example")
```

tests/testthat/test-parallel.R
```diff
@@ -147,9 +147,26 @@ test_that("can print parallel control", {
...
+    cleanup <- hipercow_example_helper(),
+    "This example uses a special helper")
+  expect_error(
+    task_create_expr(sqrt(2),
+                     resources = hipercow_resources(cores = 1),
+                     parallel = hipercow_parallel(method = "future")),
+    "You chose parallel method 'future', with 1 core")
+  expect_message(cleanup(), "Cleaning up example")
```
richfitz, line 160: then drop this bit

R/parallel.R
```diff
@@ -224,3 +224,22 @@ hipercow_parallel_setup_parallel <- function(processes, cores_per_process) {
...
+      "You chose parallel method '{parallel$method}', with 1 core",
+      "You need multiple cores for this - check your hipercow_resources"))
+  }
+
+  if (!is.null(parallel) && !is.null(parallel$cores_per_process) &&
+      parallel$cores_per_process > cores) {
+    cli::cli_abort(c(
+      paste("You chose {parallel$cores_per_process} core{?s} per process,",
+            "but requested only {cores} core{?s} in total")))
+
```
richfitz, line 241:
```suggestion
```

Lots of these blank lines creep in, I tend to drop them when I next edit the file - I've not yet worked out if there's a pattern as to where you put them though...

richfitz (changes requested):
Looking good, some minor suggestions

R/parallel.R
weshinsley: It's a bit subjective! Where I find a block of code starts to get hard to read, and there's a natural pause for breath...!

weshinsley: I made this function better, and it didn't need the blank line to look clear to me anymore...!

richfitz (approved):

PR #106, opened 2024-02-09
mrc-5058: allow unexported functions with task_create_call

richfitz:
I noticed this when porting the approach over to rrq, hopefully nothing that the users will do too much, but it can be useful


weshinsley (approved):
Looks good!

PR #108, opened 2024-02-13
Load environment on parallel workers

richfitz:
Needs bootstrap updating after merge.

Basic idea is to load an environment on the workers after creation. Default to the same environment as the worker process, but allow specification of a different one within the parallel configuration


R/parallel.R
```diff
@@ -69,31 +69,47 @@
 ##' process when launching a local cluster using one of the parallel
 ##' methods. By default, this will be `1`. See details.
 ##'
+##' @param environment The name of the environment to load into your
+##'   parallel workers.  The default is to use the environment that
+##'   you submit your task with (which defaults to `default`), which
+##'   means that each worker gets the same environment as your main
+##'   process.  This is often what you want, but can mean that you
+##'   load too much into each worker and incurr a speed or memory
```
weshinsley, line 77:
```suggestion
##'   load too much into each worker and incur a speed or memory
```

weshinsley (approved):
Looks good!

PR #111, opened 2024-02-16
Support for purging (deleting) tasks

richfitz:
Fairly big, but I think this is a nice approach.

Most of the logic is around selecting which tasks to delete.  I've gone with some basic filters (we can add more) and have intersected the results of each filter. So "failed tasks more than 3 weeks old" is something easy to delete now.  There's a bit of a fiddle with retry chains but nothing too hard. Dependent tasks, once we get them, will be another challenge.

The pkgdown index entry suggests that this might have the wrong name; we could make it `task_purge` and slot it in with the tasks, or we could call it `hipercow_cleanup` which fits within the utilities.  Or perhaps it's fine and I'm overthinking it.

The one thing not done here is cope with doing anything with the driver.  For windows there's nothing to do, but in general (and for the elsewhere driver even) we should delete the tasks remotely too.  For another PR though I think?

Also not done - links from `hipercow_bundle_delete` and from any vignette


weshinsley (commented):
I think this all looks good and makes sense! - a few thoughts on the way to maybe chat on this mornig and decide whether we merge now or do any more on this

weshinsley:
> Fairly big, but I think this is a nice approach.

Yep - I like it - I hadn't really thought through before how it affects retries and bundles...
 
> The pkgdown index entry suggests that this might have the wrong name; we could make it `task_purge` and slot it in with the tasks, or we could call it `hipercow_cleanup` which fits within the utilities. Or perhaps it's fine and I'm overthinking it.

The way this can get tangly with retries/bundles/partial success purgnig things/dependencies on the future made me feel this is something you do to your hipercow root, rather than something you do to individual tasks - so a `hipercow_` function didn't feel out of place to me. Cleanup is ok - but I think that implies to me "cleaning up of temp files/leftover debris" where we are really being destructuve here and deleting potentially useful things that you are reasonable sure you don't want. Purge feels good?

> The one thing not done here is cope with doing anything with the driver. For windows there's nothing to do, but in general (and for the elsewhere driver even) we should delete the tasks remotely too. For another PR though I think?

Yes, another PR! wpia-hn is set to 30 days at the moment, and we should at least document that in the right place - I'm not sure if there's a mechanism for manually purging

> Also not done - links from `hipercow_bundle_delete` and from any vignette

Let's make some documentation tickets...

weshinsley (approved):
All looks good!

PR #117, opened 2024-05-02
Check for problematic code in environment

richfitz:
This PR adds a check for problematic code in sources used for the environment creation.  At the moment this is limited to `install.packages()` but we might want to grow this over time (and perhaps have warnings about use of `library()` for example).


richfitz:
Not sure what is up with codecov here, frankly - I've added a test that covers those lines and while the comment has updated, the upload has not worked (despite the token).  I suspect that we are reaching the end of codecov working for us, especially with the big "pricing" banner we now see on login to it...

weshinsley (approved):
Looks good!

PR #119, opened 2024-05-10
Load environment variables at the start of the tasks' process.

plietar:
There are environment variables that control the behaviour of the R interpreter, such as tuning the GC. These variables must be set at the start at the process. Setting them from R code, as we do now, is too late and has no effect.

This commit exports all non-secret environment variables into an Renviron file in the task directory, which the batch script configures using R_ENVIRON_USER. Once of the first things the R interpreter does is load that file and set values it finds as environment variables.

The existing mechanism to load environment variables is kept. It is necessary both to support secret variables, which we don't want to expose in the Renviron, and to allow tasks to run without starting a new process at all, as is commonly done in tests and in the example driver.

Retry tasks need a bit of care for this to work as well. For the most part, the new task's folder only contains an indirection to the original task's data and not a copy of it. When the task starts, it follows the indirection to find the actual data. This strategy does not work for the Renviron file, so a copy of that file is made.

### DIFF
--- a/R/task-create.R
+++ b/R/task-create.R
@@ -531,8 +531,23 @@ task_create_bulk_call <- function(fn, data, args = NULL,
 }
 
 
+task_export_envvars <- function(root, dst, envvars, inherit_envvars) {
+  if (is.null(inherit_envvars)) {
+    envvars_export(envvars, dst)
+  } else {
+    src <- file.path(path_task(root$path$tasks, inherit_envvars), RENVIRON)
+    if (fs::file_exists(src)) {
+      fs::file_copy(src, dst)
+    } else {
+      fs::file_create(dst)
+    }
+  }
+}
+
+
 task_create <- function(root, type, path, environment, envvars,
-                        parallel, ..., id = ids::random_id()) {
+                        parallel, ..., id = ids::random_id(),
+                        inherit_envvars = NULL) {
   time <- Sys.time()
   dest <- path_task(root$path$tasks, id)
   fs::dir_create(dest)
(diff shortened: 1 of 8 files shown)


plietar:
As a manual end-to-end test, this prints `Inf` on the first task but 128 on the second one:
```
id <- task_create_expr(mem.maxVSize())
task_wait(id)
print(task_result(id))

e <- hipercow_envvars(R_MAX_VSIZE=128*1024*1024)
id <- task_create_expr(mem.maxVSize(), envvars=e)
task_wait(id)
print(task_result(id))
```

drivers/windows/inst/templates/task_run.bat
```diff
@@ -16,6 +16,7 @@ cd {{hipercow_root_path}}
 ECHO working directory: %CD%
 
 set R_LIBS_USER={{hipercow_library}}
+set R_ENVIRON_USER=hipercow\tasks\{{task_id_1}}\{{task_id_2}}\Renviron
```
richfitz, line 19: This is a great idea

R/task-create.R
```diff
@@ -531,15 +531,32 @@ task_create_bulk_call <- function(fn, data, args = NULL,
 }
 
 
+task_export_envvars <- function(root, dst, envvars, inherit_envvars) {
```
richfitz, line 534: There's a similar issue with resources on a retried task where we might want to modify and might want to inherit.  This approach seems fine for now at least

richfitz (approved):
I didn't know about that approach for setting the renviron file to read, very cool

PR #121, opened 2024-05-15
Check that versions are new enough

richfitz:
Checks on configure and provision that the versions used are not too old; too old is older than r-oldrel, at which point packages are hard to install.  We don't check this on use because at that point things should continue to work reasonably well.

The provided message is fairly gory but hopefully clear enough, and points people in the right direction. It can't be silenced easily so hopefully people will upgrade


weshinsley (approved):
Very nice!

PR #123, opened 2024-05-22
Make provisioning check optional

richfitz:


weshinsley (approved):
HMMMM.

Well, I think we should merge this and see if it helps Maggie, or whether she runs into other problems. I am perplexed, as checking the running tasks is a very sensible thing to do, and I don't understand why Maggie got the issue she did. 

In the past, we saw some behaviour like this because of excessively long (30 seconds + ) queries to the API - I forget the details, but many jobs were in the queue at the time, and the API was returning much more data than it needed to. (The data would then get filtered client-side) - it ended up being many megabytes of data to process and return on mrcdata, which caused the timeout in some of the PHP code.

I can't think of any reason why that would be happening here on the BuildQueue though. So ideally we would want to dig into this and work out what on earth is going on - my suspicions would be towards something computer-specific though - intermittent network connection / zscaler issues / something else, since we have only one report of it, and that people are provisioning all the time.

PR #124, opened 2024-05-23
Remove duplicated processing of arguments in rrq setup.

plietar:
The `hipercow_rrq_workers_submit` function was doing some processing of its resources and envvars arguments before passing them to `task_create_bulk_call`. The latter applied the exact same processing to these arguments.

For the most part, the processing is idempotent and only seeks to validate the inputs, so this only had a limited impact. The only exception is the encryption of secret variables, which was done twice, leading to undesirable behaviour.

Somehow the second encryption was failing rather than just performing a second layer of encryption. I think our use of the encryption routines has an upper limit on the size of payloads, and the outcome of the first round is larger than that limit. Regardless, we shouldn't be calling the encryption code twice, even if it had worked we would only decrypt once and not get back the correct value.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: hipercow
 Title: High Performance Computing
-Version: 1.0.18
+Version: 1.0.19
 Authors@R: c(person("Rich", "FitzJohn", role = c("aut", "cre"),
                     email = "rich.fitzjohn@gmail.com"),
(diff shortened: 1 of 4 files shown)


richfitz (approved):

richfitz:
Up to you if you want to deal with those whitespace changes but would be nice if you don't mind

plietar:
Gets me every time :disappointed:

PR #127, opened 2024-06-04
Support and docs for stan

richfitz:
~Merge after #126, contains those commits~

This PR adds a document describing the state of play for getting cmdstanr running, possibly it's too sarcastic


weshinsley (approved):
This looks good to me now - but maybe rerun from your side as a final test...

PR #128, opened 2024-06-05
Fix the path to the status file in the batch file.

plietar:
The batch file to run tasks checks for the presence of a `status-success` file when exiting, and derives its exit code from that. This is done to report useful status on the HPC web portal.

Unfortunately the path it was looking for was incorrect, therefore the file never existed and all jobs are reported as failed. This fixes the path by removing a spurious `log\` component.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: hipercow
 Title: High Performance Computing
-Version: 1.0.20
+Version: 1.0.21
 Authors@R: c(person("Rich", "FitzJohn", role = c("aut", "cre"),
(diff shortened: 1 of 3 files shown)


plietar:
I haven't actually tested this yet because the cluster is currently full :disappointed:

plietar:
Completed the testing and it seems to all work now! The job IDs are 632705, 632706 and 632707 for the success, failure and "catastrophic failure" cases.

Turns out there was a second bug, in that we need to look for the file's existence before we unmount the network drives.

richfitz:
Thanks both!

PR #131, opened 2024-06-14
mrc-5358: Report absolute path for root when helpful in hipercow_init

weshinsley:


R/root.R
```diff
@@ -24,15 +24,21 @@
 ##' path <- withr::local_tempfile()
 ##' hipercow_init(path)
 hipercow_init <- function(root = ".", driver = NULL, ...) {
+  abs_path <- fs::path_abs(root)
+  not_same_path <- !((abs_path == root) ||
```
richfitz, line 28: we depend on fs, so we might be able to use a function there to test same paths more portably and safely.

richfitz (changes requested):

R/root.R
weshinsley: fs::path_norm does what we want - (fs::path_abs is already normalised)

richfitz (approved):

PR #133, opened 2024-06-28
Print environment contents when loading it.

plietar:
hipercow silently attaches packages and sources files from the configured environment. If something goes wrong in this process (eg. the file does not exist), it can be difficult for users to understand what is happening. None of their own code makes any mention of the files or packages in question, and they may have forgotten that they ever created the environment.

This prints a message when loading the environment, with the list of files and packages that will be loaded. This should provide enough hints for the user to make the connection between a hypothetical error and the environment.

### DIFF
--- a/R/task-eval.R
+++ b/R/task-eval.R
@@ -38,7 +38,7 @@ task_eval <- function(id, envir = .GlobalEnv, verbose = FALSE, root = NULL) {
     version <- utils::packageVersion("hipercow")
     cli::cli_h1("hipercow {version} running at '{root$path$root}'")
     cli::cli_alert_info("library paths:")
-    cli::cli_li(.libPaths())
+    cli::cli_ul(.libPaths())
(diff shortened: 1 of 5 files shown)


R/task-eval.R
```diff
@@ -38,7 +38,7 @@ task_eval <- function(id, envir = .GlobalEnv, verbose = FALSE, root = NULL) {
     version <- utils::packageVersion("hipercow")
     cli::cli_h1("hipercow {version} running at '{root$path$root}'")
     cli::cli_alert_info("library paths:")
-    cli::cli_li(.libPaths())
+    cli::cli_ul(.libPaths())
```
plietar, line 41: `cli_li` was leaving a container open, which then indented all other bullet points in the function.
`cli_ul` does not have this problem, as it closes the container when given a non-NULL argument.

weshinsley (approved):
Looks good!

PR #134, opened 2024-07-05
mrc-5090 - add dry_run to task_purge

weshinsley:


R/task-purge.R
```diff
@@ -122,14 +126,16 @@ hipercow_purge <- function(task_ids = NULL,
   }
```
richfitz, line 127: Is it worth printing something indicating that the following lines are a work of fiction? We have (in the task logs printing) a pair of hoizontal rules that frame a block of output, here we might add something like that?

R/util.R
```diff
@@ -471,3 +471,18 @@ unlist_times <- function(x) {
 empty_time <- function() {
   Sys.time()[-1]
 }
+
+
+maybe_unlink <- function(x, recursive = FALSE, dry_run = FALSE) {
+  if (!dry_run) {
+    unlink(x, recursive)
+    return()
+  }
```
richfitz, line 480: I know we usually go for a happy path style, but I wonder if this one should be one big if/else?

R/util.R
```diff
@@ -471,3 +471,18 @@ unlist_times <- function(x) {
...
+  recurse <- if (recursive) " recursively."
+  cli::cli_alert_info("Dry_run - would have deleted:")
+  ul <- cli::cli_ul()
+  for (file in x) {
+    cli::cli_li("{file}{recurse}")
+  }
+  cli::cli_end(ul)
```
richfitz, line 487:
```suggestion
  cli::cli_bullets(paste0(file, recurse))
```

(this formats very slightly differently out the box but you can tame it)

richfitz (changes requested):

R/task-purge.R
weshinsley: Done with a cli_rule before and after with pointy arrows...

R/util.R
weshinsley: Done - basically
    files <- paste0(x, recurse)
    names(files) <- rep("*", length(files))
    cli::cli_bullets(files)

richfitz (approved):

PR #135, opened 2024-07-05
Mrc 5429 - warn on "..." specified for null driver in hipercow_init

weshinsley:


R/root.R
```diff
@@ -24,6 +24,16 @@
 ##' path <- withr::local_tempfile()
 ##' hipercow_init(path)
 hipercow_init <- function(root = ".", driver = NULL, ...) {
+  if ((is.null(driver)) && (length(list(...)) > 0)) {
```
richfitz, line 27: Perhaps look into `rlang::check_dots_used` (https://rlang.r-lib.org/reference/check_dots_used.html) which might do this for us?

R/root.R
```diff
@@ -24,6 +24,16 @@
...
+    args <- list(...)
+    cli::cli_alert_warning(
+      "driver was not specified, but extra args were detected: ")
+    ul <- cli::cli_ul()
```
richfitz, line 31: See comments in #134 about a better way to do these

richfitz (changes requested):

R/root.R
weshinsley: We decided to omit these...

R/root.R
weshinsley: There is the possibility of... 

if (is.null(driver)) {
  rlang::check_dots_empty()
}

but converting this from an error to a warning, and including a message to say why 

! `...` must be empty.

only applies when driver = NULL was a bit fiddly.

richfitz (approved):

PR #136, opened 2024-07-07
Print env values as well as names in task startup

weshinsley:


R/task.R
```diff
@@ -679,11 +679,17 @@ print_info_data <- function(data) {
   }
   cli::cli_li("Environment: {data$environment}")
   if (NROW(data$envvars) > 0) {
-    cli::cli_li("Environment variables: {data$envvars$name}")
+    cli::cli_li("Environment variables:")
+    env_vars_ul <- cli::cli_ul()
```
richfitz, line 683: https://cli.r-lib.org/reference/cli_dl.html perhaps, possibly as

```
cli::cli_dl(items = set_names(data$envvars$value, data$envvars$name)
```

For secrets though, you should only print stars!

richfitz (commented):
This has bothered me too, thanks. One suggestion here that will shorten the implementation, and I can't remmeber what happens with secrets at this point

richfitz (approved):

PR #140, opened 2024-07-25
Docs for workers

richfitz:
~This still needs the final vignette rebuilt before we merge~


vignettes_src/workers.Rmd
```diff
@@ -0,0 +1,287 @@
...
+
+If you have thousands and thousands of tasks to submit at once you may not want to flood the cluster with them all at once. Each task submission tends to be relatively slow on every platform that we have used, and submitting thousands of tasks will take minutes (even if you could submit 10 tasks a second, a thousand tasks will take almost two minutes, and a million tasks would take 27 hours!). Some cluster schedulers also slow down as the queue size increases, becoming less efficient at distributing work. And if you take up the whole cluster someone may come and find you in order to complain.  At the same time, batching your tasks up into little bits and manually sending them off is a pain and work better done by a computer.  An alternative is to submit a (relatively small) set of "workers" to the cluster, and then submit tasks to them.  We call this the **"lightweight queue pattern"**.
```
weshinsley, line 30:
```suggestion
If you have thousands and thousands of tasks to submit you may not want to flood the cluster with them all at once. Each task submission tends to be relatively slow on every platform that we have used, and submitting thousands of tasks will take minutes (even if you could submit 10 tasks a second, a thousand tasks will take almost two minutes, and a million tasks would take 27 hours!). Some cluster schedulers also slow down as the queue size increases, becoming less efficient at distributing work. And if you take up the whole cluster someone may come and find you in order to complain.  At the same time, batching your tasks up into little bits and manually sending them off is a pain and work better done by a computer.  An alternative is to submit a (relatively small) set of "workers" to the cluster, and then submit tasks to them.  We call this the **"lightweight queue pattern"**.
```

vignettes_src/workers.Rmd
```diff
@@ -0,0 +1,287 @@
...
+
+The second use case is where you want to run some computation on the cluster that needs to run some of its calculations in parallel, as you might do following instructions in `vignette("parallel")`.  Suppose that your cluster only has nodes with 32 cores though, and you have some calculations that would benefit from more cores than this.  Or suppose that you want to farm out some smaller number number of sub-tasks each of which will consume an entire node.  You can use the approach described here to scale your tasks off a single node, and we call this the **"interprocess communication pattern"**.
```
weshinsley, line 32:
```suggestion
The second use case is where you want to run some computation on the cluster that needs to run some of its calculations in parallel, as you might do following instructions in `vignette("parallel")`.  Suppose that your cluster only has nodes with 32 cores though, and you have some calculations that would benefit from more cores than this.  Or suppose that you want to farm out some smaller number of sub-tasks each of which will consume an entire node.  You can use the approach described here to scale your tasks off a single node, and we call this the **"interprocess communication pattern"**.
```

vignettes_src/workers.Rmd
```diff
@@ -0,0 +1,287 @@
...
+Let's submit 1,000 trivial tasks, using `rrq_task_create_bulk_expr`, taking the square root of the first thousand positive integers.
+
+```{r, include = FALSE}
+t0 <- Sys.time()
+```
+```{r}
+ids <- rrq_task_create_bulk_expr(sqrt(x), data.frame(x = 1:1000))
+```
+
+There's no equivalent of a task bundle in `rrq`; this just returns a vector of task 1000 task identifiers.  You can pass this vector in to `rrq_task_wait()` though, and then fetch the results using `rrq_task_results()` (note the pluralisation; `rrq_task_results()` always returns a list, while `rrq_task_result()` fetches a single task result).
```
weshinsley, line 143:
```suggestion
There's no equivalent of a task bundle in `rrq`; this just returns a vector of 1000 task identifiers.  You can pass this vector in to `rrq_task_wait()` though, and then fetch the results using `rrq_task_results()` (note the pluralisation; `rrq_task_results()` always returns a list, while `rrq_task_result()` fetches a single task result).
```

vignettes_src/workers.Rmd
```diff
@@ -0,0 +1,287 @@
...
+now submit a task that will call this function:
+
+```{r}
+id <- task_create_expr(
+  example(16),
+  parallel = hipercow_parallel(use_rrq = TRUE))
+id
+```
+
+Here, `id` is a *hipercow* task identifier.  This sends an additional task to the queue, so now we have five things running on the cluster (four workers and one task).  The task runs, picks up the controller, then uses it do distribute a (trivial) calculation over the four workers.
```
weshinsley, line 225:
```suggestion
Here, `id` is a *hipercow* task identifier.  This sends an additional task to the queue, so now we have five things running on the cluster (four workers and one task).  The task runs, picks up the controller, then uses it to distribute a (trivial) calculation over the four workers.
```

vignettes_src/workers.Rmd
```diff
@@ -0,0 +1,287 @@
...
+task_result(id)
+```
+
+You can see from the worker logs here the tasks being split between the workers:
+
+```{r}
+rrq_worker_log_tail(n = 32)
+```
+
+This example is trivial, but you could submit 10 workers each using a 32 core node, and then use a single core task to farm out a series of large simulations across your bank of computers.  Or create a 500 single core workers (so ~25% of the cluster) and smash through a huge number of simulations with minimal overhead.
```
weshinsley, line 238:
```suggestion
This example is trivial, but you could submit 10 workers each using a 32 core node, and then use a single core task to farm out a series of large simulations across your bank of computers.  Or create a team of 500 single core workers (so ~25% of the cluster) and smash through a huge number of simulations with minimal overhead.
```

vignettes_src/workers.Rmd
```diff
@@ -0,0 +1,287 @@
...
+
+## Controlling the worker environment
+
+The workers will use the `rrq` environment if it exists, failing that the `default` environment.  So if you need different packages and sources loaded on the workers on your normal tasks, you can do this by creating a different environment
+
+```{r}
+hipercow_environment_create("rrq", packages = "cowsay")
+```
+
+**TODO**: *work out how to refresh this environment; I think that's just a message to send*
```
weshinsley, line 252: Is this ok for now, or shall we solve it?

vignettes_src/workers.Rmd
```diff
@@ -0,0 +1,287 @@
...
+
+You can submit your workers with any resources and parallel control you want (see `vignettes("parallel")` for details); pass these as `resources` and `parallel` to `hipercow_rrq_workers_submit()`.
+
+
+# General considerations
+
+## Stopping redundant workers
+
+By default, workers will live for 10 minutes after finishing their last task.  This means that most of the time that you use workers you can largely forget about cleanup.  If you want be polite and give up these resources early (this would be important if you wanted to launch new workers, or if you were using a large fraction of the cluster), you can tell them to stop immediately after completion of the last job by setting their timeout to zero.
```
weshinsley, line 261:
```suggestion
By default, workers will live for 10 minutes after finishing their last task.  This means that most of the time that you use workers you can largely forget about cleanup.  If you want to be polite and give up these resources early (this would be important if you wanted to launch new workers, or if you were using a large fraction of the cluster), you can tell them to stop immediately after completion of the last job by setting their timeout to zero.
```

weshinsley (approved):
Typos for a redocument - and one TODO that I'm not immediately sure I know what to do with - which we could do, or leave...

PR #141, opened 2024-08-16
Use an absolute path for R_LIBS_USER and R_ENVIRON_USER.

plietar:
At present, the R_LIBS_USER path embedded in the batch file is relative to the working directory in which R is spawned. This generally works fine.

However, if the user's task switches working directories and then spawns a new R interpreter, then the new session will not be able to resolve the R_LIBS_USER anymore and the user's provisioned library won't be found.

This was an issue when a user was running an orderly task, which switches working directories to a draft folder, and then uses the parallel package, which spawns new R processes. Note that this is only an issue when users create their own clusters and not when using the builtin hipercow parallel support, since the latter explicitly inherits the libPaths.

By making the path absolute it now works regardless of which working directory the new R processes are spawned.

This also changes R_ENVIRON_USER for the same reasons. You could argue that sub-processes should not reload that file anyway, since they inherit the environment variables directly already. However if we did want that behaviour it should be done separately and consistently, probably by unsetting R_ENVIRON_USER, and not just happen to be the case when the working directory is modified.

### DIFF
--- a/drivers/windows/R/paths.R
+++ b/drivers/windows/R/paths.R
@@ -68,7 +68,7 @@ print.windows_path <- function(x, ...) {
 
 remote_path <- function(x, shares) {
   x <- prepare_path(x, shares)
-  windows_path_slashes(file.path(x$path_remote, x$rel, fsep = "/"))
+  file.path(x$drive_remote, x$rel)
 }
 
 
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: hipercow
 Title: High Performance Computing
-Version: 1.0.33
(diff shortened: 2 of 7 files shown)


drivers/windows/R/paths.R
```diff
@@ -68,7 +68,7 @@ print.windows_path <- function(x, ...) {
 
 remote_path <- function(x, shares) {
```
plietar, line 69: This function wasn't being used anywhere outside of its own unit tests, so I've re-used the name and changed its semantics a bit to fit what I wanted.

weshinsley (approved):
All looks good to me - keen for Rich's view too

PR #145, opened 2024-10-18
Configure the rrq offload directory each time a controller is created.

plietar:
We previously setup the rrq offload directory once when creating the queue, storing the absolute path to it in the redis database. Using a single global value breaks down when that directory is mounted on multiple machines but at differing paths.

The most spectacular failure case happens when configuring the queue from a linux machine and using it to run on a Windows cluster. When this happens, the offload path looks something like
`/home/<user>/path/to/project/hipercow/rrq/offload`, which when used from a Windows machine obviously does not work.

This is solved by configuring each instance of an `rrq_controller` with its own path to the offload directory, relative to the hipercow root. This root is already automatically adjusted based on where we are running.

### DIFF
--- a/R/rrq.R
+++ b/R/rrq.R
@@ -179,7 +179,9 @@ rrq_prepare <- function(driver, root, offload_threshold_size,
   timeout_idle <- 300 # 5 minutes
   heartbeat_period <- 60 # one minute
 
-  r <- rrq::rrq_controller(queue_id, con, offload_threshold_size = offload_threshold_size, ...)
+  r <- rrq::rrq_controller(queue_id, con,
+                           offload_threshold_size = offload_threshold_size,
+                           ...)
 
   cfg <- rrq::rrq_worker_config(timeout_idle = timeout_idle,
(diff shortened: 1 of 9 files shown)


plietar:
This requires https://github.com/mrc-ide/rrq/pull/125 to be released first.

It will also require a bootstrap library update with the new versions of both rrq and hipercow.

R/rrq.R
```diff
@@ -33,12 +32,30 @@ hipercow_rrq_controller <- function(..., set_as_default = TRUE, driver = NULL,
                                     queue_id = NULL, root = NULL) {
   root <- hipercow_root(root)
   call <- rlang::current_env()
+
```
richfitz, line 35: Perhaps `loadNamespace` with `versionCheck` here to make sure we have `0.7.20` or later

richfitz (approved):

R/rrq.R
plietar: Used `compareVersion` instead for nicer error messages.

PR #150, opened 2024-11-25
List tasks

richfitz:
We get asked for this fairly frequently and there's no real reason not to implement it.  This PR adds a basic `task_list`, using the machinery developed for `task_purge`.  The actual code is very short, but the PR looks a bit larger than it is because I moved the code around a bit so that the names make sense

### DIFF
--- a/R/task-list.R
+++ b/R/task-list.R
@@ -123,8 +123,8 @@ task_select_by_status <- function(task_ids, with_status, root, call) {
   }
 
   ## This is actually quite hard to do efficiently, and might
-  ## benefit from a progress bar.  If one of the statues that we're
-  ## interesed in is *not* 'created' we can probaly just look for
+  ## benefit from a progress bar.  If one of the statuses that we're
+  ## interesed in is *not* 'created' we can probably just look for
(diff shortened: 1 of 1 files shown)


R/task-list.R
```diff
@@ -0,0 +1,195 @@
...
+    return(task_ids)
+  }
+
+  if (!is.null(task_ids)) {
+    i <- task_status(task_ids, root = root, follow = FALSE) %in% with_status
+    return(task_ids[i])
+  }
+
+  ## This is actually quite hard to do efficiently, and might
+  ## benefit from a progress bar.  If one of the statues that we're
```
weshinsley, line 126:
```suggestion
  ## benefit from a progress bar.  If one of the statuses that we're
```

R/task-list.R
```diff
@@ -0,0 +1,195 @@
...
+  ## interesed in is *not* 'created' we can probaly just look for
```
weshinsley, line 127:
```suggestion
  ## interesed in is *not* 'created' we can probably just look for
```

weshinsley (approved):

weshinsley (approved):

PR #152, opened 2024-11-26
Add docs on environments

richfitz:
I've been meaning to write this for ages.  There's still some details to add about parallel tasks, but that will open its own can of worms so leaving that for now

### DIFF
--- a/vignettes/environments.Rmd
+++ b/vignettes/environments.Rmd
@@ -1,0 +1,243 @@
+---
+title: "Environments"
+output: rmarkdown::html_vignette
+vignette: >
+  %\VignetteIndexEntry{Environments}
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteEncoding{UTF-8}
+---
+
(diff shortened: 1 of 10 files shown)


vignettes/environments.Rmd
```diff
@@ -0,0 +1,243 @@
...
+
+```{r}
+id <- task_create_expr(length(my_big_data))
+task_wait(id)
+task_result(id)
+```
+
+If you use this approach, you should be aware of a few things:
+
+* There is no guarantee that the object you see when creating the task is the same as the object loaded by `sources`.  We can add validation if you would find that useful
```
richfitz, line 209: this might actually be a lie and is controlled by `hipercow.validate_globals` https://mrc-ide.github.io/hipercow/articles/details.html#hipercow-validate_globals

richfitz: ok, updated now

weshinsley (approved):

PR #153, opened 2024-11-26
Refresh worker environment after update

richfitz:
After we update the `rrq` environment we should send an environment refresh request to any active workers, otherwise people will be confused that their tasks don't use the new environment (this is noticed in the docs already in #140).

This does mean that if a user creates the rrq controller, creates the rrq env and then launches workers, the workers will load the environment twice but that's fine.

### DIFF
--- a/vignettes/parallel.Rmd
+++ b/vignettes/parallel.Rmd
@@ -19,11 +19,6 @@ cleanup <- withr::with_dir(
                                     new_directory = FALSE,
                                     initialise = FALSE))
 library(hipercow)
-
-```
-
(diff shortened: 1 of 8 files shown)


vignettes/parallel.Rmd
```diff
@@ -9,6 +9,7 @@ vignette: >
 
 ```{r setup, include = FALSE}
 source("common.R")
+options(hipercow.timeout = 120)
```
weshinsley, line 12: Probably don't need this any more, but it doesn't matter either

weshinsley (approved):

PR #154, opened 2024-11-27
Mrc 6077 - Fix failure with 500G memory resource request

weshinsley:


drivers/windows/R/web.R
```diff
@@ -271,12 +271,12 @@ client_body_submit <- function(path, name, resources, cluster,
 
   if (!is.null(resources$memory_per_node)) {
     req$mpn <- encode64(as.character(
-      1000 * resources$memory_per_node))
+      1000L * resources$memory_per_node))
```
richfitz, line 274: I'd prefer `format(x, scientific = FALSE)` here, particularly because you risk conversion to NA here if we overflow integer size.  So for this case I think this is

```suggestion
    req$mpn <- encode64(format(
      1000 * resources$memory_per_node),
       scientific = FALSE))
```

or you may wish to add a utility

```r
as_character_integer <- function(x) {
  format(x, scientific = FALSE)
}
```

and use that?

weshinsley: Yep, let's do the latter...

richfitz (approved):

PR #155, opened 2024-11-27
Enable Testing queue

weshinsley:
Bit bodgy - the section this touches should really be handled by querying the API, instead of the templates being hard-coded into hipercow. But for today...

### DIFF
--- a/drivers/windows/R/resource.R
+++ b/drivers/windows/R/resource.R
@@ -13,7 +13,7 @@ cluster_info_wpia_hn <- function() {
   resources <- list(
     max_cores = 32,
     max_ram = 512,
-    queues = c("AllNodes", "Training"),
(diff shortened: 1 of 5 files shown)


drivers/windows/R/resource.R
```diff
@@ -13,9 +13,9 @@ cluster_info_wpia_hn <- function() {
   resources <- list(
     max_cores = 32,
     max_ram = 512,
-    queues = c("AllNodes", "Training"),
+    queues = c("AllNodes", "FastQueue"),
```
richfitz, line 16: I wonder if we should use `Testing` and not `FastQueue` here, to emphasise that the queue is for testing work?

weshinsley: Yeah, good call - perhaps "FastQueue" has other surface-level implications too...

PR #156, opened 2024-11-28
Add information on the bulk interface

richfitz:
Another long-standing omission; not really sure how people are finding how to do this!


weshinsley (approved):
This all looks good - some little suggestions (and perhaps we could later on do a bigger cancel/delete section if that seems worth it, but not today.)

Afraid I didn't get any further with the insane CI problems...

PR #158, opened 2024-11-28
Try fixing temp location

richfitz:
~Should do the same thing as the previous fix, but should work? https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs~

That didn't work, but this somehow does. Failure is just codecov adding noise, per usual


weshinsley (approved):
Splendid

PR #160, opened 2024-11-29
Another approach to better temporary directories

richfitz:
This is used in #140 and #156, unfortunately it's not a complete fix as I'm still seeing some failures


weshinsley (approved):
Oops, did these out of order - already seen in the bundle docs PR

PR #161, opened 2024-12-06
Better username msgs

weshinsley:


drivers/windows/R/dide_auth.R
```diff
@@ -23,6 +23,13 @@ windows_authenticate <- function(call = NULL) {
 
   username <- check_username(
     readline_with_default("DIDE username", windows_guess_username()))
+  if (grepl(" ", username)) {
```
richfitz, line 26: Use `[# ]` perhaps, and again in the error?

drivers/windows/R/dide_auth.R
```diff
@@ -23,6 +23,13 @@ windows_authenticate <- function(call = NULL) {
...
+    cli::cli_abort(
+      c("Usernames must not contain spaces.",
```
richfitz, line 28: Change this perhaps to reflect the user error and not the Auth error. So "The username you have provided does not look valid as it contains {invalid}" with some logic to report back the specific complaint

weshinsley: Try now?

richfitz (approved):

PR #163, opened 2025-01-23
mrc-6185 - Linux support on Win driver part 1

weshinsley:
A bit gnarly - and hopefully a good place to stop. Functionally nothing should change with this PR; there is no actual linux-handling yet, but it prepares the way by taking a `platform` argument to various functions, and storing a few things as a list per-platform as required. For now all the extra platform args are set to `windows` by default, so the tests also don't change too much, except where they specifically look at the args.

Main hipercow: these functions now allow `platform` to be specified - the idea being these are done at a time where we haven't set resources for the resources/queue a task should run on, so we could be thinking about either platform here

hipercow_hello
hipercow_provision
hipercow_provision_list
hipercow_provision_check
hipercow_provision_compare
hipercow_resources_validate

hipercow_configure - I've left to handle this through the existing `...` for now, since not all drivers will support multiple platforms - which makes me wonder if `...` might not be better for the 5 functions above - I'm not quite sure. But have to start somewhere. 

Then in hipercow.windows, the places where `platform` gets added flow naturally from the above. Actual code changes in hipercow.windows:-

* `config$r_versions` and `config$path_lib` used to be a vector of versions and a string respectively; each is now a list of potentially two of the above, named `windows` and `linux`. It's not a very nice idea but I think we need to be prepared to have different available r_versions on the different platforms.
* We also need, I think, a web_client per platform, with the platform name and other platform-specific things as private properties. web_client gets cached, so we now have a list of two web clients cached - the reuslts of the cluster_resources and r_versions api-calls also get cached, and are now lists too.
* Places that use config$path_lib then follow - so the building of the package path, including the bootstrap library. These do not exist yet, but the code will expect (or create) `bootstrap-linux` and `bootstrap-linux-dev` folders, with R versions within, alongside the "default" existing win versions.
* And the creating of the task from the template require path_lib, so changes in there to swith on platform...

At this point, I _believe_ we have something still consistent with no changes for windows users, but all the scaffolding to start acting on when a different platform is specified...

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: hipercow
 Title: High Performance Computing
-Version: 1.0.53
+Version: 1.0.54
 Authors@R: c(person("Rich", "FitzJohn", role = c("aut", "cre"),
                     email = "rich.fitzjohn@gmail.com"),
              person("Wes", "Hinsley", role = "aut"),
--- a/R/cluster.R
+++ b/R/cluster.R
@@ -7,8 +7,8 @@
 ##' @param driver The driver to use, which determines the cluster to
 ##'   fetch information from (depending on your configuration).  If no
 ##'   driver is configured, an error will be thrown.
-##'   
-##' @param platform The operating system you are intending to run 
+##'
+##' @param platform The operating system you are intending to run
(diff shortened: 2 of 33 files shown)


weshinsley:
Reimplemented better on mrc-6185-2

PR #164, opened 2025-01-24
Quickfix on ints from API resources call

weshinsley:


drivers/windows/R/web.R
```diff
@@ -392,7 +392,10 @@ client_parse_r_versions <- function(txt) {
 
 
 client_parse_cluster_resources <- function(txt) {
-  from_json(txt)
+  dat <- from_json(txt)
+  dat$max_cores <- as.integer(dat$max_cores)
```
richfitz, line 396: I think this means that the json being emitted from the web api is incorrect? are you sure you don't want to fix that there?

weshinsley: Yes, you're completely right...!

PR #168, opened 2025-02-20
Hipercow Linux Support

weshinsley:


drivers/windows/DESCRIPTION
```diff
@@ -29,9 +29,11 @@ Imports:
     keyring,
     logwatch,
     openssl,
+		pkgdepends,
```
richfitz, line 32: can you make sure your editor is not set to use tabs and fix this?

drivers/windows/NAMESPACE
```diff
@@ -7,3 +7,4 @@ S3method(format,dide_shares)
 S3method(print,dide_clusterload)
 S3method(print,password)
 S3method(print,windows_path)
+importFrom(utils,read.csv)
```
richfitz, line 10: Any reason not to use `utils::read.csv` on use?

drivers/windows/R/batch.R
```diff
@@ -1,19 +1,48 @@
 write_batch_task_run <- function(task_id, config, path_root) {
+
+  run_on_linux <- config$platform == "linux"
+  template_file <- if (run_on_linux) "task_run.sh" else "task_run.bat"
+  out_script <- if (run_on_linux) SH_RUN else BATCH_RUN
+  write_os_lines <- if (run_on_linux) write_linux_lines else writeLines
```
richfitz, line 6: This might be easier to read in one if/else

```
if (run_on_linux) {
  template_file <- ...
  out_script <- ...
} else {
}
```

etc

drivers/windows/R/batch.R
```diff
@@ -1,19 +1,48 @@
...
+  path <- path_to_task_file(path_root, task_id, out_script)
+  write_os_lines(str, path)
+
+  if (run_on_linux) {
+    path_dat <- prepare_path(path, config$shares)
+    linux_path <- path_on_linux(path_dat)
+    wrap_path <- path_to_task_file(path_root, task_id, SH_WRAP_RUN)
+    write_linux_lines(
+      sprintf("python -u /opt/hpcnodemanager/kwrap.py %s", linux_path),
+        wrap_path)
```
richfitz, line 19: Can you add a comment explaining what any of this does please?

drivers/windows/R/batch.R
```diff
@@ -1,19 +1,48 @@
...
+  }
   path
 }
 
 
 write_batch_provision_script <- function(id, config, path_root) {
+  run_on_linux <- config$platform == "linux"
+  template_file <- if (run_on_linux) "provision.sh" else "provision.bat"
```
richfitz, line 27: The other way of doing this is you have a list 

```
details <- list(
  linux = list(template_file = "provision.sh", write_lines = write_linux_lines),
  windows = list(template_file = "provision.bat", write_lines = writeLines)
```

and then you do

```
template_file <- details[[config$platform]][["template_file"]]
```

or similar.  But I think the version above is probably easiest tbh

drivers/windows/R/batch.R
```diff
@@ -30,13 +59,14 @@ template_data_task_run <- function(task_id, config, path_root) {
   data$task_id_2 <- substr(task_id, 3, nchar(task_id))
 
   data$hipercow_library <- paste(
-    remote_path(file.path(path_root, config$path_lib), config$shares),
+    remote_path(file.path(path_root, config$path_lib), config$shares,
+                platform == "linux"),
```
richfitz, line 63: why are we passing a boolean here for linux rather than just `platform`? The latter seems more general and means we don't have to think about if we're dealing with it booleanised or not

drivers/windows/R/batch.R
```diff
@@ -60,12 +90,17 @@ template_data_common <- function(config, path_root) {
     "ECHO Removing mapping {{drive}}\nnet use {{drive}} /delete /y",
     network_shares_data)
 
+  r_version <- config$r_version
+  if (config$platform != "linux") {
+    r_version <- version_string(config$r_version)
+  }
```
richfitz, line 96: What is going on here? Why a conversion into string only for not linux?

drivers/windows/R/batch.R
```diff
@@ -60,12 +90,17 @@ template_data_common <- function(config, path_root) {
...
+
   list(
     hostname = hipercow:::hostname(),
     date = as.character(Sys.time()),
     hipercow_version = hipercow_version(),
     hipercow_windows_version = hipercow_windows_version(),
-    r_version = version_string(config$r_version),
+    r_version = r_version,
```
richfitz, line 103: This change seems odd, given the odd change above? Half way through refactoring?

drivers/windows/R/batch.R
```diff
@@ -80,10 +115,11 @@ path_bootstrap <- function(config) {
   version <- version_string(config$r_version, ".")
   if (platform == "windows") {
     ## TODO: update to I:/bootstrap(-dev)?/(windows|linux)/<version>
+    ## - Bit of a pain to migrate as I:/bootstrap is active.
+    ## - Can we tolerate I:/bootstrap, I:/bootstrap-dev  and
+    ##   /wpia-hn/Hipercow/bootstrap-linux /wpia-hn/Hipercow/bootstrap-dev-linux
```
richfitz, line 120: We'll need to fix this very soon, I think.  We can start with the new version saving things into -windows pretty soon. Can you make a ticket and we'll do this in the next few days after merging this

drivers/windows/R/bootstrap.R
```diff
@@ -1,31 +1,68 @@
-bootstrap_update <- function(development = NULL, root = NULL) {
+bootstrap_update <- function(development = NULL, root = NULL,
+                             platform = "windows") {
```
richfitz, line 2: move platform first and drop default arg

drivers/windows/R/bootstrap.R
```diff
@@ -1,31 +1,68 @@
...
   path_script <- "hipercow/bootstrap-windows.R"
   path_root <- hipercow:::hipercow_root(root)$path$root
   path_script_abs <- file.path(path_root, path_script)
   dir.create(dirname(path_script_abs), FALSE, TRUE)
   bootstrap <- read_template("bootstrap.R")
+  prefix <- if (platform == "windows") "I:" else "/wpia-hn/Hipercow"
+  suffix <- if (platform == "windows") "" else "-linux"
```
richfitz, line 9: group as suggested before. But I do think that we possibly should just move immediately to to `-windows` too at which point the suffix calculation is very easy

drivers/windows/R/bootstrap.R
```diff
@@ -1,31 +1,68 @@
...
+
+  bootstrap_repos <- c("https://mrc-ide.r-universe.dev",
+                       "https://cloud.r-project.org")
+  bootstrap_repos <- sprintf('c("%s")',
+                             paste0(bootstrap_repos, collapse = "\", \""))
+
+  bootstrap_pkgs <- c("hipercow", "remotes", "pkgdepends", "renv", "rrq")
+  deps <- pkgdepends::pkg_deps$new(bootstrap_pkgs)
+  deps$resolve()
+  bootstrap_pkgs <- deps$get_resolution()$package
```
richfitz, line 19: can you pull this bit into a small function

drivers/windows/R/driver-linux.R
```diff
@@ -0,0 +1,46 @@
+hipercow_driver_linux <- function() {
+  linux_driver <- hipercow_driver_windows()
```
richfitz, line 2: We will regret this. Can you please split into dide_driver_base() and then a linux and a windows version that add their relevant bits

drivers/windows/R/mounts.R
```diff
@@ -1,3 +1,5 @@
+##' @importFrom utils read.csv
```
richfitz, line 1: I don't think we want this generally, but if we do it should be above the function that uses `read.csv` not the top of the file. But just tweak the function in question to quieten the note

drivers/windows/R/mounts.R
```diff
@@ -178,3 +180,46 @@ dide_locally_resolve_unc_path <- function(path, mounts = detect_mounts(),
...
+
+  remap2 <- function(path_dat, server, folder, dest) {
+    path_to_folder <- sprintf("\\\\\\\\%s\\\\%s\\\\", server, folder)
+    if (grepl(path_to_folder, path_dat$path_remote)) {
+      inner_folder <- gsub(path_to_folder, "", path_dat$path_remote)
+      return(sprintf("/%s/%s/%s", dest, inner_folder, path_dat$rel))
+    }
+    FALSE
+  }
+
```
richfitz, line 202: Can this be documented and/or refactored so that it's more obvious what it is doing please?

drivers/windows/R/mounts.R
```diff
@@ -178,3 +180,46 @@ dide_locally_resolve_unc_path <- function(path, mounts = detect_mounts(),
...
+  try_wpiahn2 <- remap(path_dat, "wpia-hn2.dide.ic.ac.uk", "wpia-hn2")
+  if (!isFALSE(try_wpiahn2)) return(try_wpiahn2)
+
+  try_wpiahn2 <- remap(path_dat, "wpia-hn2.hpc.dide.ic.ac.uk", "wpia-hn2")
+  if (!isFALSE(try_wpiahn2)) return(try_wpiahn2)
+
+  cli::cli_abort(c(
+    "Error mapping linux path",
+    i = "Couldn't work out linux mount for {path_dat$path_remote}"))
+}
```
richfitz, line 225: Sorry, this really needs docs to explain what is going on.

drivers/windows/R/paths.R
```diff
@@ -66,9 +66,13 @@ print.windows_path <- function(x, ...) {
 }
 
 
-remote_path <- function(x, shares) {
+remote_path <- function(x, shares, run_on_linux = FALSE) {
```
richfitz, line 69: see above, this should be platform

drivers/windows/R/paths.R
```diff
@@ -66,9 +66,13 @@ print.windows_path <- function(x, ...) {
...
   x <- prepare_path(x, shares)
-  file.path(x$drive_remote, x$rel)
+  if (run_on_linux) {
+    path_on_linux(x)
```
richfitz, line 72: this name I think is insufficiently clear as to what sort of changes are going on here, and in particular I think that this is really "path on the dide network as seen from a linux machine", as opposed to some sort of windows/linux slash normalisation that one might expect

drivers/windows/R/web.R
```diff
@@ -36,10 +36,10 @@ web_client <- R6::R6Class(
     },
 
     submit = function(path, name, resources, cluster = NULL,
-                      depends_on = NULL) {
+                      depends_on = NULL, workdir = "") {
```
richfitz, line 39: What is going on with workdir here? I'm not thrilled that we need to set it on linux but not windows and that is has a default -- which presumably won't work on linux

drivers/windows/R/web.R
```diff
@@ -109,11 +109,12 @@ web_client <- R6::R6Class(
...
     },
 
-    cluster_resources = function(cluster, driver) {
-      r <- private$client$GET(
-        sprintf("/api/v1/cluster_info/%s/%s", cluster, driver),
-        public = TRUE)
-      client_parse_cluster_resources(httr_text(r))
+    cluster_resources = function() {
+      r_windows <- private$client$GET(
+        "/api/v1/cluster_info/wpia-hn/hipercow.windows/", public = TRUE)
```
richfitz, line 114: These names are immediately out of date as they look like package names (we had `hipercow.windows` but now we are working towards `hipercow.dide`.  I know that the portal will be replaced by a new API soon but we might want to not bake in these names all the same

drivers/windows/inst/templates/task_run.sh
```diff
@@ -0,0 +1,54 @@
...
+export PATH=/opt/apps/lmod/lmod/libexec:$PATH
+source /opt/apps/lmod/lmod/init/bash
+export LMOD_CMD=/opt/apps/lmod/lmod/libexec/lmod
+module use /modules-share/modules/all
+
+module load R/{{r_version}}
+module load hiredis
+module load libsodium
+
+echo working directory: $(pwd)
```
richfitz, line 18: in the windows version the first bit of the script moves to the working directory. Is there some reason this won't work in linux? If not then we probably should move that bit of control into the same place on windows too

drivers/windows/inst/templates/task_run.sh
```diff
@@ -0,0 +1,54 @@
...
+echo running on: $(hostname -f)
+
+export PATH=/opt/apps/lmod/lmod/libexec:$PATH
+source /opt/apps/lmod/lmod/init/bash
+export LMOD_CMD=/opt/apps/lmod/lmod/libexec/lmod
+module use /modules-share/modules/all
+
+module load R/{{r_version}}
+module load hiredis
+module load libsodium
```
richfitz, line 16: what ends up needing libsodium?

richfitz (changes requested):
The biggest concern I have is that the linux platform support is still to obviously done by difference to the windows one. Other small comments throughout

drivers/windows/DESCRIPTION
weshinsley: Done - missed that file from last commit...

drivers/windows/NAMESPACE
weshinsley: Fixed

drivers/windows/R/batch.R
weshinsley: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-6247/Bootstrap-locations

drivers/windows/R/mounts.R
weshinsley: Done

drivers/windows/R/batch.R
weshinsley: Refactored into batch-linux.R and batch-windows.R with batch.R doing the common stuff, so hopefully this looks nicer now.

drivers/windows/R/batch.R
weshinsley: Comments added in batch-linux.R

drivers/windows/R/batch.R
weshinsley: Hopefully nicer with batch-linux refactor

drivers/windows/inst/templates/task_run.sh
weshinsley: Oh - that's a mistake - it's hipercow.windows itself that needs it (keyring), but that's not needed on the node...

drivers/windows/R/batch.R
weshinsley: Done. I might go a bit further, since having three occurrences of config$ on this line looks a bit overkill

drivers/windows/R/mounts.R
weshinsley: Done

drivers/windows/R/paths.R
weshinsley: Renamed.

drivers/windows/R/paths.R
weshinsley: Done

drivers/windows/R/mounts.R
weshinsley: Done hopefully

drivers/windows/R/bootstrap.R
weshinsley: Done

drivers/windows/R/bootstrap.R
weshinsley: Done

drivers/windows/R/bootstrap.R
weshinsley: Refactored

drivers/windows/R/batch.R
weshinsley: Provisionally gone for option 1 on that ticket - we'll talk tomorrow...

drivers/windows/R/batch.R
weshinsley: The version_string converts "4.4.2" into "4_4_2" which windows wants for its "call setr64_4_4_2.bat", but linux does not, for its "module load R/4.4.2"

I'll make this more clear.

drivers/windows/R/batch.R
weshinsley: r_version at this point is the correct thing for either OS.

drivers/windows/R/batch.R
weshinsley: Commented

drivers/windows/R/web.R
weshinsley: Maybe `windows` and `linux` are fine here, to match config$platform

richfitz (changes requested):
Pop some light testing over some of the batch file bits and some decent testing on the unc path rewriting, and then we can patch the rest up later

drivers/windows/R/batch.R
```diff
@@ -1,20 +1,9 @@
...
-
 write_batch_provision_script <- function(id, config, path_root) {
-  data <- template_data_provision_script(id, config, path_root)
-  str <- glue_whisker(read_template("provision.bat"), data)
-  path_job <- file.path(path_root, "hipercow", "provision", id)
-  path <- file.path(path_job, "provision.bat")
-  fs::dir_create(path_job)
-  writeLines(str, path)
-  path
+  if (config$platform == "windows") {
```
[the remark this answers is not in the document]
weshinsley: Done

drivers/windows/R/web.R
weshinsley: I've updated to use windows and linux

drivers/windows/tests/testthat/test-batch.R
```diff
@@ -85,3 +85,14 @@ test_that("can write a provision batch file", {
...
 })
+
+
+test_that("Can prepare a linux task", {
+  mount <- withr::local_tempfile()
+  root <- example_root(mount, "b/c")
+  path_root <- root$path$root
+  config <- root$config[["dide-linux"]]
+  id <- "12345678123456781234567812345678"
+  #res <- write_batch_task_run_linux(id, config, path_root)
```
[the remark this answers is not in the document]
richfitz: this comment was from a previous review through, please disregard :)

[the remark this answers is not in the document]
weshinsley: Cool - yes, I distinctly remember half-finishing that one!! then thinking "I'll try a different angle and test the inner bits first"

PR #169, opened 2025-03-10
hipercow.windows -> hipercow.dide

weshinsley:
So this ugly PR accomplishes...

1. Rename hipercow.windows to hipercow.dide in various places.

2. In main hipercow, rename windows.R to dide.R, and various function renaming.

3. In hipercow.dide, various functions starting with `windows_` were in driver-windows.R, but were shared between windows and linux. Renamed these to `dide_` and moved them to driver-base.R. 

4. Push through tests and vignettes.

NB - github is showing quite a confusing diff here - particularly:-
* it thinks windows/driver-windows.R has been morphed into dide/driver-base.R - I can sort of understand why, as much of driver-windows.R was common, and was moved into driver-base.R , but it looks more messy than it is, and a manual diff may be clearer

Also, tests/testthat/test-windows.R has been renamed to test-dide.R with some changes; github diff has not picked that up very well, so test-dide.R is appearing as a new file.

To-Do - regenerate vignettes from vignettes_src

Lots more doc work to do, especially around the "Windows" focussed one - but for another PR.


R/configuration.R
```diff
@@ -77,21 +77,27 @@ configuration_paths <- function(root) {
...
+        dide_username(),
+        error = function(e) {
+          if (!message_written) {
+            cli::cli_warn(
+                   c("Failed to read username",
+                     i = "Try 'dide_username()' to reproduce separately"),
+                 parent = e)
+            message_written <- TRUE
+          }
         "(???)"
```
richfitz, line 98: move into its own function perhaps, with a `warn` argument

richfitz (changes requested):

richfitz (approved):

PR #172, opened 2025-03-28
Ignore deprecated 'windows' driver if 'dide-windows' exists

weshinsley:


R/configure.R
```diff
@@ -271,6 +271,17 @@ hipercow_driver_select <- function(name, required, root, call = NULL) {
         c("No hipercow driver configured",
           i = "Please run 'hipercow_configure()' to configure a driver"),
         call = call)
+
+    } else if ((length(valid) == 2) &&
+             setequal(valid, c("windows", "dide-windows"))) {
```
richfitz, line 276: This can be simplified to

```suggestion
    } else if (setequal(valid, c("windows", "dide-windows"))) {
```

I think, because the first test is part of the second

weshinsley: Yes, true

richfitz (approved):

